Issue #36, opened 2023-07-09
Returning typesafe erros


M-jerez:
At the moment he managing the errors throwing an error from the handlers.
But this lost the type safety when errors are thrown. so we probably need to return the error types ro alway return a router error

* merge PublicError and RouteError into a single Class RpcError, that does not serialize the message, just the public message.
* make it generic so status code, name and errorData, can be specified statically. ie:
```ts
Class RpcError <statusCode, name, errorData> { ...
```
* investigate if error name should be handled a an unique code

M-jerez:
### Copying this from PR as this issue will be kept open!

Maybe we could create a new `ApplicationError<StatusCode, Name>` and infer the status code and name from the types when creating the error using reflection. 

Also if we add the error to the return type, then on the  client we will need to create a type map to remove the errors from the return type and throw them so they are handled by the client in the cath method.

Server:
```ts
export const addTwo= (value:any): number | ApplicationError<400, 'Not A Number'> => {
    if (isNan(value)) {
        throw new ApplicationError<400, 'Not A Number'>('value is not a number');
    }
    return value + 2;
};
``` 

Client:
```ts
type AddTwo = typeof addTwo;
type Success = AnythinButErrors<AddTwo>; // type = number
type Fail = OnlyErrors<AddTwo>; // type = ApplicationError<400, 'Not A Number'>

   response
   .then((sum: Success) => {
        // ... do anything
    })
   .catch((err: Fail) => {
        // ... do anything
   });
``` 


This las thing of types Promise error is unknown in typescript, so maybe is a good selling point for the client.

**All this is quite a bit of rewort so maybe prioritise for later.**

M-jerez:
Some more work has been done into this regard. https://github.com/MionKit/mion/pull/53

Each public hook/route now returns a tuple [`returnValue`,`RouteError`]. `RouteError` is undefined when the request is succesful and `returnValue` is undefined when the request fails.
```ts
{
  <routeName>: [routeResponse, RouteError],
  <hookName>: [hookResponse, RouteError],
}
```
* Returned Value = response.body[<routeName>][0]
* Returned Error = response.body[<routeName>][1]

PR #39, opened 2023-07-09
manage returning errors instead throwing them

M-jerez:
Partial work on  #36.

Thi PR adds support for returning errors instead throwing them. but returned errors are still generic.

To add typesafe errors The error code and name should be at least generic.

Maybe we could create a new `ApplicationError<StatusCode, Name>` and infer the status code and name from the types when creating the error using reflection. 

Also if we add the error to the return type, then on the  client we will need to create a type map to remove the errors from the return type and throw them so they are handled by the client in the cath method.

Server:
```ts
export const addTwo= (value:any): number | ApplicationError<400, 'Not A Number'> => {
    if (isNan(value)) {
        throw new ApplicationError<400, 'Not A Number'>('value is not a number');
    }
    return value + 2;
};
``` 

Client:
```ts
type AddTwo = typeof addTwo;
type Success = AnythinButErrors<AddTwo>; // type = number
type Fail = OnlyErrors<AddTwo>; // type = ApplicationError<400, 'Not A Number'>

   response
   .then((sum: Success) => {
        // ... do anything
    })
   .catch((err: Fail) => {
        // ... do anything
   });
``` 

This las thing of types Promise error is unknown in typescript, so maybe is a good selling point for the client.

**All this is quite a bit of rewort so maybe prioritise for later.**

### DIFF
--- a/packages/http/src/mionHttp.ts
+++ b/packages/http/src/mionHttp.ts
@@ -19,7 +19,7 @@ import {DEFAULT_HTTP_OPTIONS} from './constants';
 import type {HttpOptions, HttpRawServerContext} from './types';
 import type {IncomingMessage, RequestListener, Server as HttpServer, ServerResponse} from 'http';
 import type {Server as HttpsServer} from 'https';
-import type {Obj, Headers, RawRequest, PublicError, RouterOptions, SharedDataFactory} from '@mionkit/router';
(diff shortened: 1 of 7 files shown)


M-jerez:
This is already done, probably as part of some other PR, so closing it without merging!

https://github.com/MionKit/mion/blob/74d505b096e518790bedda928d3223bb53773dee/packages/router/src/dispatch.ts#L145-L166

PR #48, opened 2023-07-18
41 using internal hooks

### LINKED ISSUE #41: Add option to wrapp already generated routes
Add a methods to prepend and append hooks.

these hooks are intended for internal router functioning. the goal is that the router only contains logic about creating and dispatching routes, and any other logic related to the request/response parsing serialized etx is handled using hooks.

This is so the router or any other plugging can add hooks without  modifying existing routes
Prepend and append hooks should be called  called before any register routes so there is no need to remade all the routes.

M-jerez:
Add a methods to prepend and append hooks.

these hooks are intended for internal router functioning. the goal is that the router only contains logic about creating and dispatching routes, and any other logic related to the request/response parsing serialized etx is handled using hooks.

This is so the router or any other plugging can add hooks without modifying existing routes
Prepend and append hooks should be called called before any register routes so there is no need to remade all the routes.

resolves #41


M-jerez:
Some work was done to implement this, it required quite a bit changes, creating new pacakges and in general adding complexity. There was quite a bit of degradation as well!

Branch here: https://github.com/MionKit/mion/tree/41-using-internal-hooks

Screenshot 2023-07-17 at 00 00 05
Screenshot 2023-07-17 at 00 00 17

Do to this we are not proceeding with the merging as it has too many disadvantages.
The idea is good so we might want to revisit in the future again. removing the async context will help simplifying things!

PR #130, opened 2026-08-01
docs(todos): upstream CompiledFnArgs fix landed; record the mion cleanup

M-jerez:
Docs only. Updates the existing `docs/todos/upstream-compiledfnargs-type-lie.md` now that the upstream half is fixed in [MionKit/ts-run-types#308](https://github.com/MionKit/ts-run-types/pull/308).

I updated the existing todo rather than filing a new one — it already tracked exactly this, and its "Done when" was *"when it ships: delete `toWireArgs`"*.

## Upstream did NOT take either fix we proposed, and the reason matters

We suggested (1) widen `defaultParamValues` to `unknown` values, or (2) drop both fields as derivable from `familyTag`. Both were wrong.

`defaultParamValues` is consumed when a receiver rebuilds the function via `new Function(...)` — which is precisely why it was typed as strings. The Go emitter's `ArgSpec` says so outright:

```go
// Key is the conceptual slot ("vλl", "pλth", "εrr"); Name is the JS identifier
// in the emitted signature; Default is the JS-source default expression
// (empty for no default).
type ArgSpec struct{ Key, Name, Default string }
```

Both tables are **JS source fragments**, describing the emitted `function verr_<hash>(v, pth=[], er=[])`. So `CompiledFnArgs` was correct all along — the JS mirror in `entryTuple.ts` had drifted to storing real runtime values where source text belongs. Upstream kept the type and corrected the values (`{vλl:'', pλth:'[]', εrr:'[]'}` and friends), deleted all five `as unknown as` casts, and documented the contract.

## Which corrects one of our own claims

Our todo said *"nothing reads these values, so the loss is inert"* and called them derivable bookkeeping. **That was wrong in principle.** They're the default expressions a client needs to reconstruct a signature. No outage resulted — our client restores through `code` via `buildFactoryFromCode` — but the field is meaningful and must be forwarded verbatim.

## What `toWireArgs` does, in the order it bites

1. **On the current pin (0.11.0) it ships wrong data.** Non-string values are dropped, so `pλth: []` and `εrr: []` vanish; then `if (!('vλl' in out)) out.vλl = 'v'` injects the *identifier* `'v'` into a table of *default expressions*. Every error-shaped entry goes over the wire as `defaultParamValues: {vλl: 'v'}` — three slots reduced to one, and that one semantically wrong.
2. **After the upgrade it becomes an identity no-op.** Every value is a string, so the filter copies all of them and the `vλl` fallback is unreachable.

## Blocked on a release

`@ts-runtypes/core` is pinned to `0.11.0` in `client`, `core`, `drizze`, `examples` and `router`. The fix is on ts-run-types `main`, unreleased — so nothing is actionable in code yet. The todo records the follow-up: bump the five pins, delete `toWireArgs`, spread `defaultParamValues` through like `args` already is, and add a round-trip assertion.

One trap noted in the todo for whoever writes that assertion: `JSON.stringify` silently drops undefined-valued keys and `toEqual` treats a missing key and an undefined one as equal, so a naive round-trip check passes *even with the bug present*. Assert on slot presence instead.

https://claude.ai/code/session_01EnD2KeUxrXbuKRcUrE9zje

---
_Generated by [Claude Code](https://claude.ai/code/session_01EnD2KeUxrXbuKRcUrE9zje)_

### DIFF
--- a/docs/todos/upstream-compiledfnargs-type-lie.md
+++ b/docs/todos/upstream-compiledfnargs-type-lie.md
@@ -1,72 +1,96 @@
 # Upstream: `CompiledFnArgs` cannot describe its own values
 
-**Status:** todo — file against `ts-run-types`, not fixable in mion
-**Type:** upstream bug report
+**Status:** todo — upstream half FIXED, mion half blocked on a release
+**Type:** upstream bug report → now a mion cleanup
 **Created:** 2026-07-27
+**Updated:** 2026-08-01 — upstream fix landed as
+[MionKit/ts-run-types#308](https://github.com/MionKit/ts-run-types/pull/308)
 **Found while:** deleting `rtResolver.ts` ([../done/rtresolver-removal.md](../done/rtresolver-removal.md))
 
-## The bug
+## The bug (as filed, and correct on `@ts-runtypes/core` 0.11.0)
 
-`@ts-runtypes/core` 0.11.0, `packages/ts-runtypes/src/runtypes/types.ts:142-147`:
+`CompiledFnData` types **both** `args` and `defaultParamValues` as `CompiledFnArgs`
+(all-string). `args` genuinely is all-strings. `defaultParamValues` is not, for any family —
+`{vλl: undefined}`, `{vλl: undefined, pλth: [], εrr: []}`, `{vλl: undefined, θpts: {}}`,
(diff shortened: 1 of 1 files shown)


M-jerez:
not needed anymore, and assumptions are wrong

PR #170, opened 2026-08-29
SPIKE: `.format()` column modifier as alternative to `refineTableType`

M-jerez:
This is a spike implementation exploring a new API for constraining column types in drizzle tables. Instead of using `refineTableType(table, {...})` to add format constraints after the fact, columns can now chain `.format({...})` directly in the builder.

## Summary

Adds a `.format()` method to the column chain that merges extra type-format params into a column's captured ones. This is measured as 29-35% cheaper than `refineTableType` in type instantiation cost, especially for downstream consumers reading models from emitted `.d.ts` files.

## Changes

- **Recorder** (`packages/drizzle-orm/src/recorder.ts`): Added type-only `format()` method that returns `this` (no runtime recording, since drizzle has no counterpart).

- **Type columns** (`packages/drizzle-orm/src/typeColumns.ts`): Exported `Format<P>` marker interface for the type-road equivalent, which merges params via `MergeFormat` instead of replacing.

- **PG columns** (`packages/drizzle-orm-pg-core/src/columns.ts`): Added `format()` signature to all four kind interfaces, reusing existing `MergeFormat` and `RefinableParamsOf` machinery.

- **Behavior tests** (`packages/drizzle-orm-pg-core/src/formatModifierSpike.spec.ts`): Pinned semantics: chains in any position, merges when called twice, stacks with `refineTableType`, compile errors match `refineTableType`'s contract, runtime is a genuine no-op.

- **Type cost measurement** (`packages/type-budget/test/formatModifierSpike.compile.test.ts`): Measured against real packages; `.format()` saves 32-35% on table declarations and 29% on downstream consumers.

- **Query interop tests** (`packages/type-budget/test/queryModelInterop.test.ts`): Pinned how `toDrizzle` synthesizes configs from either view; refined rows don't round-trip to unrefined models when refinements overwrite captured params (e.g., `min` on integers).

- **Column refinability audit** (`packages/type-budget/test/columnRefinability.test.ts`): Enumerated all pg columns; found 8 that map incorrectly (uuid, cidr, macaddr, interval, sparsevec, numeric/decimal in default mode).

## Implementation notes

- Reuses `MergeFormat` / `RefinableParamsOf` from `@ts-runtypes/core`, no new type-level concepts.
- Runtime is a no-op; `toDrizzle()` produces byte-identical drizzle configs whether the table was formatted or refined.
- Type-road `Format<P>` marker costs ~410 instantiations per column (same band as builder `.format()`), but cannot validate param families at declaration time (needs `NormalizeCol` work, not in prototype).
- Recommendation: ship BOTH as the default way to constrain a column, keep `refineTableType` for the one case it's still needed (a second view of the same table with different rules).

## Status

Spike result — the idea works and is measurably cheaper. Nothing is built yet beyond the pg prototype this document measures. All tests pass; declaration emit is clean.

https://claude.ai/code/session_01GazkkGRht2kD9ZwCPLgN3k

### DIFF
--- a/docs/maybe/drizzle-column-format-modifier.md
+++ b/docs/maybe/drizzle-column-format-modifier.md
@@ -87,6 +87,8 @@ did not show up in the consumer measurement.
 
 ## Semantics, all verified on the prototype
 
+Behaviour is pinned by `packages/drizzle-orm-pg-core/src/formatModifierSpike.spec.ts`.
+
 - Chains in any position: before `notNull()`, after it, before `default()`.
 - Chains twice, and the two merge (`.format({minLength: 10}).format({maxLength: 20})`).
 - `refineTableType` still stacks on top of a formatted table, and wins on a shared key.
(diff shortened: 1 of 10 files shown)


M-jerez:
Closing: the `.format()` idea is discarded. The direction is to stay aligned with the drizzle API and split the domains instead, so a modifier that has no drizzle counterpart is the wrong shape.

The spike answered its question and the diff stays readable here for the record. Three findings in it are about SHIPPED behaviour, not about `.format()`, so they outlive this PR:

1. **Format params are invariant.** `String<{maxLength: 50}>` is not assignable to `String<{maxLength: 100}>` in either direction, yet the same move via the base compiles in two steps. Reproduces in `@ts-runtypes/core` with no drizzle in the program. Nothing pins format-to-format assignability today; `typesafety.test.ts` only pins format vs its base.

2. **Refined rows do not always round trip.** A refinement that ADDS a param key leaves the row assignable back to the unrefined model; one that OVERWRITES a captured key (`{age: {min: 18}}` over `Int32`'s own `min`) does not. Numeric columns hit this on every refinement.

3. **Eight pg columns cannot be refined at all.** `uuid` carries the `uuid` format but that family is missing from `RefinableParamsByFamily`, so `{version: '4'}` is rejected. `cidr`, `macaddr`, `macaddr8`, `interval`, `sparsevec` and `numeric`/`decimal` in their default mode map to a bare `string`, so they carry no params to merge into. `cidr` is the odd one, since `inet` maps to `IP`.

---

PR #183, opened 2026-08-30
Close silent divergence on standard-library types per lib

M-jerez:
## Summary

Fixes the silent failure where a standard-library type's shape differs across TypeScript lib editions, causing the same source to compile to different validators with no diagnostic. The root cause was an incomplete blocklist of type names; the fix inverts the logic to decide data by what IS data instead.

## Key changes

- **Closed the silent-divergence hole** by replacing the `NonSerializableGlobals` name list with a predicate that tests whether a type is declared in the bundled standard library. A lib-declared interface or class now projects atomically (no members walked) regardless of its name or what edition added it.

- **Added `LibDeclaredGlobalOf`** (new file `libglobal.go`) to detect standard-library declarations by checking the symbol's declaration location, not its name. This handles the finite set of "what IS data" instead of trying to enumerate "what is NOT data".

- **Guarded against empty lib selections** with `CFG002` diagnostic. When `lib: []` or no base ECMAScript edition is selected, TypeScript never declares `Array`, so `number[]` resolves to an empty object and the validator would silently accept anything. The resolver now refuses this configuration outright.

- **Added `LibSet` type** (new file `libset.go`) to read what standard library the Program actually loaded from its source files, accounting for reference chains and implicit defaults. This replaces re-deriving the lib from tsconfig spelling.

- **Removed the name lists** (`NonSerializableExactGlobals`, `NonSerializableBaseGlobals`) and the `MKR014` catch-all diagnostic that tried to warn when the list fell behind. The closed contract means the set can now grow deliberately instead of types sneaking in.

- **Updated test coverage** with new suites pinning the behavior: `libatomic_test.go` verifies lib types project whole, `libguard_test.go` verifies CFG002 catches unsound lib selections, and `libdrop_test.go` verifies the per-family drop warnings are announced not silent.

- **Added TypeScript floor notice** (`typescript-floor.ts`) to warn at build time when the consumer's TypeScript is below the supported floor (6.0.0), since an older version can write a tsconfig the resolver reads differently.

## Notable details

- Binary types are still recognized by shape (`ArrayBufferView` member pattern) rather than by name, so user subclasses of typed arrays work without a list.

- Declaration merging is safe: if any declaration of a lib interface sits outside the lib directory, it's treated as the author's data and walked normally.

- Type aliases can never be caught by the lib test (they have no interface/class flag), so `Partial<T>`, `Record<K, V>`, etc. stay data even when materialized from a lib alias.

- The structural ID of a lib type is now stable across lib editions, fixing the case where `Intl.DateTimeFormat` had four different IDs depending on which lib was selected.

https://claude.ai/code/session_012EGeVjzaMTC2bvjLxKoM4a


container/website/sites/runtypes/content/02.guide/03.validation.md
```diff
@@ -9,9 +9,40 @@ Every `createX<T>()` factory turns a type into a purpose-built function the buil
...
+  link: URL;
+}
+
+// after
+interface Bookmark {
+  title: string;
+  link: string; // url.href
+}
+```
+
```
M-jerez: no need for this sction, this is an implementation detail, no public facing

M-jerez: Agreed, removed in 70d8415.

One thing I kept, because dropping the section would otherwise lose information the page had before this PR: the note above used to name binary values (`Uint8Array`, `Buffer`) among what gets skipped, and I had moved that into the section. It now sits back in the note as one clause, with no mechanism:

> Non-serializable members (functions, methods, getters and symbols) are skipped, along with built-in values that have no JSON form, such as `URL` or a typed array. None of them survive a round trip through JSON or the wire anyway, and you get a build warning naming each one that is dropped.

Shout if you would rather that clause went too.

---

PR #187, opened 2026-08-31
feat(runtypes): checkUnknowns fuses the unknown-key check into validators

M-jerez:
Strict validation costs two compiled functions run back to back today:

```ts
const isUser = createValidateFn<User>();
const hasExtraFast = createHasUnknownKeysFn<User>(undefined, {runsAfterValidation: true});
const isUserStrict = (d: unknown): d is User => isUser(d) && !hasExtraFast(d);
```

Two walks of the value, every nested object visited twice. `createValidateFn` and `createGetValidationErrorsFn` now take `{checkUnknowns: true}` and emit ONE function that checks properties and undeclared keys in a single pass:

```ts
const isUserStrict = createValidateFn<User>(undefined, {checkUnknowns: true});
```

## Two families, not a variant

This is the load-bearing design decision. Compile-time variants are **root-scoped**: `renderEntryWithDeps` keeps the plain family's `InnerPrefix`, so a NAMED nested type is dependency-called on the plain entry and its extra keys go unchecked. Variants are also never disk-cached (both the read and the write in `module.go` are gated on `variantSuffix == ""`) and they skip user overrides.

So `checkUnknowns` selects new operations (`vst` / `vest`) rather than a variant. Families render their own transitive subtree, disk-cache under their own tag, and honour overrides. `TestCheckUnknowns_RendersFusedEntryForNestedNamedType` is the regression guard.

The call site's marker still says `'val'` / `'verr'` — the scanner swaps the operation in `computeSiteFn`, so no marker type changed and the fused families inherit `noLiterals` / `numberMode` / `rejectCircularRefs` unchanged.

## Placement is the win

Each fused emitter embeds its plain twin and overrides nothing; the only difference is a key check spliced into the shared object arms. It is appended **after** the property checks, which is what makes the O(1) `countEnumKeys` compare sound at every depth — every declared prop is known present by the time it runs. That is exactly the precondition standalone callers must promise via `runsAfterValidation`, and can only honour at the root. Optional-prop and non-RT shapes fall back to the key-list scan; index-signature shapes take no check at all, since every key matching the index IS declared.

**One predicate decides whether a node takes a check at all.** `emitsUnknownKeyCheck` owns every gate — strict family, callable shape, index signature, no declared names — and both fused arms are a single `if` on it. They used to spell the conditions out separately and agree by coincidence, which is how two of the bugs below got in.

## Behaviour worth reviewing

**Error order.** `verr(v).concat(uke(v))` groups every type error ahead of every unknown-key error. One walk cannot produce that grouping, so the fused report **interleaves per node in walk order**, like every other error family. Documented and pinned.

**Unions answer per branch.** This is the one shape where the fused form does not equal `validate(v) && !hasUnknownKeys(v)`, and it is a choice. `hasUnknownKeys` never validates, so it cannot know which member matched; it pools every member's property names into one allowlist. The fused validator inherits validate's OR chain, so each arm carries its own key check:

```ts
type Pet = {kind: 'cat'; meows: boolean} | {kind: 'dog'; barks: number};
const mixed = {kind: 'cat', meows: true, barks: 3};

isPet(mixed) && !hasUnknown(mixed); // true  — barks is declared somewhere
isPetStrict(mixed);                 // false — barks is not declared on Cat
```

The fused answer follows the branch `isType` matched, so it is the one that ships. A key belonging to NO member is rejected by both. The error form reports `{path: [], expected: 'union'}` for a union, since the key is only undeclared relative to a branch. Documented in the option's doc comment, the shipped spec and the website guide, and pinned by test.

## One rule: an array is never key-checked

An array cannot carry undeclared object properties — its enumerable keys ARE its elements. `arraySkipsKeyCheck` is now the single place that rule lives, rendering the skip in whichever algebra the caller composes in, and **all four families go through it**: the blind predicate, its `runsAfterValidation` variant, `validateStrict` and `validationErrorsStrict`. Four families that must answer alike, one helper, so they cannot drift.

The object guard splits rather than disappearing, and both halves are emitted **exactly once**:

- The `typeof` / `!== null` half is redundant once validation has run, and the fused validator emits it itself as the leading term of the same chain. So the key check must not repeat it.
- The array half is NOT redundant. Passing validation does not prove a value is not an array, because an array can structurally satisfy an object shape (`[1, 2]` is a `{length: number}`, `['x']` is a `{0: string}`).

An ARRAY-TYPED node emits no key check at all, only the traversal, and each element carries its own — `[{a: 'x', evil: 1}]` against `Item[]` is still rejected. Pinned from both directions.

## FnHashLen 3 → 4, and the bug it caused

Registering the two operations tripped the collision guard: `"validationErrors|NAM"` and `"validateStrict|~C"` both hashed to `"Vtr"`. Bumping is the remedy the guard prescribes; renaming an operation to dodge a collision is what it forbids. The cost is a one-time churn of every emitted cache key, which invalidates on-disk caches and the generated TS mirror. Consumers resolving through `getFnHash` (mion's `JIT_FUNCTION_IDS`) picked the new values up with no edit.

**The bump broke something first.** `FN_HASH_LEN` in `entryTuple.ts` is a hand-written mirror of the Go constant with no drift guard. Bumping only the Go side left the JS side slicing 4-char hashes to 3:

```ts
if (runTypeId !== undefined) key = key.slice(0, FN_HASH_LEN) + '_' + runTypeId;
```

That builds a key nothing is registered under, the lookup misses, and the factory degrades to the family noop — a validator returning `true` for every input. Only the value-first form (`createValidateFn(schema)`) rebuilds that key, so the type form kept working and hid it completely. Every hand-written test passed; the fuzz lane caught it. `getFnHash.test.ts` now pins the two constants together.

## Review round: four defects the corpus could not reach

An independent review drove real compiled functions through shapes the suite had none of. The oracles were sound; the corpus was too narrow, so these sat under a green suite.

1. **Strict errors ignored unknown keys inside a union.** The union error arm hard-coded the PLAIN validate entry, so `validateStrict` rejected `{kind:'circle', radius:1, extra:true}` while `validationErrorsStrict` reported `[]` — a caller asking why got nothing.
2. **The array rule was inconsistent across the four families.** Settled as above.
3. **The unknown-key families never descended into a union member.** `{tag:'n', inner:{x:1, evil:2}}` came back clean from both `hasUnknownKeys` and `unknownKeyErrors`. Every unambiguous merged property is now walked. A property two members declare with different object shapes stays skipped: picking a branch without validating is what this family refuses to do, and guessing would report the other branch's keys on a clean value.
4. **A type error in the strict suite** that `pnpm run lint` catches but a narrower run does not.

The corpus grew to match: object unions (discriminated and not), a callable shape, an index signature, a Map/Set, and two array-satisfiable shapes — in the strict suite, the fuzz targets and the Go emit tests, plus a disk-cache round trip under the `vst` / `vest` tags.

## Tests

- `check_unknowns_test.go` — routing, emit shape (count compare vs scan vs no check), option composition, the nested-named-type case, the union arms, the callable branch, the array rule from both directions, every guard term emitted exactly once, both families gating alike, and the Marker coverage rule.
- `checkUnknowns.test.ts` — behaviour, a parity oracle against the two-call composition, the error-order pin, unions of named interfaces, and the array block including descent through arrays, arrays held by a property, and tuples.
- `test/suites/strict-validation/` — the case table, five oracles per case. A union cannot satisfy the parity pair, so it carries an explicit divergence assertion instead: at least as strict as the composition on every sample, strictly stricter on at least one.
- **O18 `fused-agree`** — the fused validator against `validate(v) && !hasUnknownKeys{runsAfterValidation}(v)`, the composition it actually replaces. **O21 `strict-self`** — the fused validator and its error twin must agree, which is what catches a rejection with an empty report. A new `extras` phase injects undeclared keys, without which both oracles ran green while proving nothing about their own subject. Both verified non-vacuous: each fires on a deliberately reverted emitter.

Full Go suite green; `pnpm run test:ci` 9024 passing; a 60s fuzz soak clean; lint, typecheck, format and codegen clean.

## Delegated findings

Two bugs surfaced by the review round reproduced on `main` and were fixed in parallel sessions.

- MionKit/mion#191 — a validate VARIANT plus a union made `getValidationErrors` contradict `validate`. **Merged.** This branch is rebased on it, and the two fixes compose: the union arm now resolves `validateStrict` under the caller's own variant.
- MionKit/mion#192 — `hasUnknownKeys{runsAfterValidation}` answering wrongly for an array. **Closed unmerged**, and folded into the one-rule change above instead: the rule turned out to span all four families, and landing half of it separately is how they drift apart again.

## Docs

Spec moved to `docs/done/fused-validators-check-unknowns.md`, rewritten to record what actually shipped, including the union and array stances and the two design reversals that contradict the original plan. Phase 3 split out as `docs/todos/create-parse-fn.md`. Website guide, a new compiled example for the union case, and the option's doc comment updated.


M-jerez:
## Benchmark: fused `checkUnknowns` vs the two-call composition

The fused function is a clear win on every shape measured, roughly 1.5x to 1.8x.

| case | fused | two calls | speedup |
|---|---|---|---|
| validate, realworld order | 12.11 M ops/s | 6.62 M ops/s | **1.83x** |
| validate, flat all-required | 77.20 M ops/s | 49.33 M ops/s | **1.56x** |
| getValidationErrors, realworld order | 2.69 M ops/s | 1.57 M ops/s | **1.71x** |

What is being compared:

```ts
// fused
const isStrict = createValidateFn<Order>(undefined, {checkUnknowns: true});

// two calls
const isOrder = createValidateFn<Order>();
const hasUnknown = createHasUnknownKeysFn<Order>();
const isStrict = (v: unknown) => isOrder(v) && !hasUnknown(v);
```

Worth noting the two rows measure different emit paths. `flat all-required` takes the O(1)
key-count compare (`cntEK(v) === 3`), `realworld order` has an optional key so it drops to the
key-array scan. The bigger win on the order case is the one that matters, because that is the
shape most real payloads have: the scan is the expensive half, and fusing it into the property
walk is what removes the second pass over every nested object.

### Method

Throwaway script, not committed. 300k iterations per measurement, best of 5 rounds (a single
timed loop showed the same comparison at 3.99x and 1.88x on consecutive runs, which is
scheduler noise). Warm-up of 30k before timing. Both sides built outside the timed loop.
Node v26.8.1, 4-core Intel Xeon @ 2.80GHz, values already valid so neither side short-circuits
early.

---

M-jerez:
## Correction to the benchmark above

The numbers in my previous comment were measured badly. Both sides ran against a single constant
input with the result discarded, so V8 was free to hoist the whole call out of the timed loop, and
whichever side it hoisted more of looked faster for no reason.

Redone with a pool of 64 distinct values indexed per iteration, and every result accumulated into
a sink. The conclusion holds, the magnitudes are smaller:

| case | fused | two calls | speedup | previously claimed |
|---|---|---|---|---|
| validate, realworld order | 20.56 M ops/s | 11.89 M ops/s | **1.73x** | 1.83x |
| validate, flat all-required | 80.08 M ops/s | 52.19 M ops/s | **1.53x** | 1.56x |
| getValidationErrors, realworld order | 6.20 M ops/s | 4.24 M ops/s | **1.46x** | 1.71x |

The error row moved most, so treat 1.46x as the number for that one.

---

PR #189, opened 2026-08-31
feat(runtypes): createParseFn decodes and checks in one call

M-jerez:
Getting a typed, validated value out of parsed JSON took three compiled functions and hand-written glue:

```ts
const restored = restoreFromJson(data);
if (!isUser(restored)) throw new Error(...getValidationErrors(restored));
```

`createParseFn<T>()` is those three in one:

```ts
const parseUser = createParseFn<User>();
const user = parseUser(JSON.parse(body)); // typed, or throws RTParseError
```

Input is the output of `JSON.parse`, not a string, so it composes with whatever read the envelope rather than duplicating `createJsonDecoderFn`.

> Rebased onto `main` now that [#187](https://github.com/MionKit/mion/pull/187) has merged, so the diff is just `createParseFn`.

## How it works

The body is a **composition of the families that already exist**, not a walk of its own. `ukuw` / `rj` / `val` / `vst` each stand for that family's cached entry for this type:

```js
function prs(v){
  v = ukuw?.fn(v) ?? v;                                  // strip strategy only
  try{ v = rj?.fn(v) ?? v }catch(e){ throw utl.parseMismatch(v,e) }
  if(!(val?.fn(v) ?? true)) throw utl.parseMismatch(v);   // vst for the fail strategy
  return v
}
```

Each `?.fn` reaches a separately compiled, separately cached, separately tuned entry, so every improvement to `validate` or `restoreFromJson` reaches parse for free and parse can never drift behind the families it duplicates. `ParseEmitter.EmitDependencyCall` panics on purpose: `Emit` renders the whole body at the root and never calls `CompileChild`, so reaching it would mean the design had quietly changed.

**This is the second design, and the first is worth knowing about** because the reasoning outlived it. See the two comments below for the full measurement trail; the short version is that a fused per-node walk ran at 0.42x of a plain `validate` while doing the same work, against 0.73x for the two calls it was meant to beat. The cause is the shape of the emitted code, not the idea: a walk that records a verdict emits statements with per-node status writes, where `validate` emits one `&&` expression the engine folds.

| | M ops/s | vs validate |
|---|---|---|
| `validate` alone | 93.9 | 1.00x |
| `restoreFromJson` + `validate` (the composition) | 68.6 | 0.73x |
| `parse{preserve}`, per-node walk | 39.6 | 0.42x |
| `parse{preserve}`, composition | 66.0 | 0.70x |

The composition reaches 0.96x of the two-call form it replaces; the remaining 4% is the one extra call frame.

**Failure is signalled by throwing.** There is nowhere else to put the verdict, since the return value is the restored data. The `ParseMismatch` sentinel carries that restored value, which is what makes `RTParseError.issues` identical to `getValidationErrors(restore(v))`, so anything already rendering those needs no second code path. Building the report is the caller's cold path, so a matching value never pays for it. An earlier design threaded a caller-provided status holder instead; it forced every consumer of the raw parse function to allocate and reuse a holder, and the allocation landed on the happy path.

## Totality

The restore arms in `json_restore.go` assume validated input: `BigInt(v)` throws a `SyntaxError`, `Temporal.X.from(v)` a `RangeError`, and the RegExp arm indexes `.match()` output with no null check. Parse is the one family whose job is untrusted input, so the whole restore call is wrapped and any throw becomes the same mismatch signal. A junk suite covers every restoring leaf; nothing may escape as anything but `RTParseError`. When `isNoopForRestoreJson` proves the restore is identity over the subtree, the call and its `try` are omitted outright.

**`issues` is never empty, and `cause` carries the raw throw.** Usually the report explains itself, because a restore arm throws precisely when the leaf is still in wire form, which is what the validator then flags at that leaf's path. A union breaks that: its members decode through an indexed envelope, so a bare `{n:'nope'}` for `{n: bigint | string}` (rather than the `{n:[0,'nope']}` the encoder writes) makes the decode throw while the undecoded value still satisfies the `string` member. The report falls back to `{path: [], expected: 'jsonForm'}` rather than nothing, and the union arm's own message reaches the caller through `cause`. Found in review, fixed in 86b0008.

Also worth flagging in review: **the no-plugin fallback throws** where every other family degrades to identity. A parse that quietly accepts everything is a hole where the caller asked for a gate.

## Undeclared keys

`strategy` picks the family, and **`preserve` is the default** — the cheapest shape (no pre-pass, no key check), and what zod does, which strips only under `.strict()`.

| strategy | composes | undeclared keys |
|---|---|---|
| `preserve` (default) | `rj` + `val` | kept |
| `strip` | `ukuw` + `rj` + `val` | blanked before the restore walks the declared shape |
| `fail` | `rj` + `vst` | rejected in one fused pass (`vst` is `validate{checkUnknowns}`, from #187) |

Strategies are separate **operations** rather than an axis: `DemandFor`'s `AxisJsonStrategy` arm assumes a composite, and `variantKey` keys entries off option names with no strategy slot, so a walking family with a strategy axis would have needed plumbing in both. Three operations needed none.

A project can set the default for every parser at once, merged site-wins exactly like `validate.numberMode`:

```ts
runtypes({parse: {strategy: 'strip'}})
```

Because parse is `AxisNone`, that default is threaded into `computeSiteFn` and resolves which family serves the site, rather than merging into an option set.

## Two bugs the tests caught

**Nested stripping silently did nothing.** A named nested type is dependency-called, and the result has to be assigned back — a parse body rebuilds rather than mutates, so a discarded result keeps the original object, extras and all. An inline nested object worked, which is what made it easy to miss. The composition removed the possibility: there are no dependency calls left to forget to assign.

**Unions do not fuse, deliberately.** The first cut assumed an `[index, payload]` envelope. The O20 round-trip oracle rejected it on the first run: `buildFlatLayout` decides *per type* whether members round-trip raw (nothing enveloped, restore is identity) or need the flat tuple, whose object branch is keyed `-1` over a merged-property layout both sides derive together. Re-deriving that here would be a second implementation of the trickiest wire format in the repo, kept in lockstep by hand. The composition made the question moot: every type, union or not, delegates to the shipped entries.

## Tests

- `internal/compiler/resolver/parse_test.go` — the body composes rather than walks, the restore call is wrapped and forwards its cause, it is omitted when nothing restores, `fail` checks through the fused validator, `strip` runs the pre-pass, failure is signalled by throwing, and the Marker coverage rule (both `getRunTypeId` shapes, paired, plus hash equivalence).
- `internal/compiler/resolver/parse_strategy_default_test.go` — the project-wide default fills in per site, an explicit `preserve` opts back out, and no default leaves every site loose.
- `packages/ts-runtypes/test/features/parse.test.ts` and `test/suites/parse/` — all three strategies, the report matching `getValidationErrors`, round-trip, the junk suite plus a sweep that every junk case reports at least one issue, the `jsonForm` fallback and its `cause`, and the bare call pinned separately from `preserve` so a change to the default cannot pass by only updating the option-carrying case.
- **O19 `parse-agree`** — two-sided against `restoreFromJson + validate`. One-sided it would have been satisfied by a parse that rejected everything; made two-sided it immediately found that the parse guards were stricter than the restore arms they guarded. **O20 `parse-round-trip`** — parse must read back whatever the JSON encoder wrote, and it caught the union bug above on its first run.

Go suite clean; `pnpm test` 9043 + 1852 passed, 0 failed; lint, typecheck and format clean.

## Docs

New section in the JSON serialization guide with a compiled example, plus a pointer from the validation page. Spec moved to `docs/done/create-parse-fn.md`, rewritten to record what shipped including both design reversals.


M-jerez:
## Benchmark: `createParseFn` vs `restoreFromJson` + `validate`

Honest result: **parse is not a speed win.** It is level with the composition when there is
something to restore, and slower when there is not.

| case | fused parse | restore + validate | ratio |
|---|---|---|---|
| Event (Date + bigint members) | 1.16 M ops/s | 1.29 M ops/s | **0.90x** |
| Order (all scalars, nothing to restore) | 16.37 M ops/s | 47.54 M ops/s | **0.34x** |
| strip vs preserve (same type) | 1.11 M ops/s | 1.13 M ops/s | 0.98x |

The Order row is the one to read carefully, and it is not really a two-call comparison. When a
type has no member that changes form on the wire, `restoreFromJson` has nothing to do, so the
"composition" collapses to a single `validate` call. Parse still walks and rebuilds, so it is
paying for a fusion that has nothing to fuse.

Where there IS something to restore (Event), the two land within 10% of each other. The
remaining gap is structural rather than incidental: `validate` compiles to one `&&` expression
the engine optimises tightly, while parse emits statement form and threads a status object so
it can record a mismatch without unwinding.

Also notable: **`strip` costs essentially nothing over `preserve`** (0.98x). The per-node object
rebuild that drops undeclared keys is not measurable next to the walk itself, which is the
reason `strip` can stay the default on security grounds without a performance argument against it.

So parse earns its place on ergonomics and safety, not throughput:

- one call instead of three, with the typed value or a throw
- **totality**: every restoring leaf is guarded, so malformed input can only ever produce
  `RTParseError`, never a raw `SyntaxError` from `BigInt()` or a `TypeError` from the RegExp arm
- undeclared keys dropped by default

I have not tried to close the gap in this PR. It would mean reworking the emitter away from
statement form, which is a different change with its own risk, and is worth doing on its own
merits rather than folded in here.

### Method

Throwaway script, not committed. 300k iterations per measurement, best of 5 rounds. Parse
restores in place so each iteration needs a fresh value; both sides allocate it from the same
cheap object-literal factory (228 M ops/s on its own, so not a factor). An earlier run used
`structuredClone` and the clone dominated both sides, hiding the difference entirely.
Node v26.8.1, 4-core Intel Xeon @ 2.80GHz.

---

M-jerez:
## Correction, and a real performance fix (4f961ff)

My benchmark above was measured badly and the conclusion it led to was wrong. Two faults: a single
constant input with the result discarded, so V8 hoisted the pure validators out of the timed loop;
and for the Order type `restoreFromJson` emits an EMPTY entry, so the "two-call" reference was
really just `validate`.

Redone with a pool of 64 distinct values and accumulated results, the gap was real but for a
fixable reason. Reading the emitted code showed parse was doing validate's job in a much worse
shape:

```js
// val  — one short-circuiting expression, no allocation
Number.isFinite(v.customer.id)

// prsp — two branches for the same check, plus a status write
if (v.customer.id === undefined) { st.ok = false; }
else { if (!(Number.isFinite(v.customer.id))) { st.ok = false; } }
```

### Two fixes

**1. One status object per compiled function, not per call.** `createParseFn` allocated a fresh
`{ok:true}` on every call. The compiled body is synchronous and never re-enters the closure, so it
is reset instead:

```ts
const status: ParseStatus = {ok: true};
return (value) => { status.ok = true; /* … */ };
```

**2. A subtree with nothing to restore delegates to the validate entry.** When the restore is
provably identity over a subtree, parse now emits one call to that subtree's `val` entry rather
than its own per-node walk. Measured as fast as inlining validate's expression outright. Preserve
only: strip must rebuild each object to drop undeclared keys and fail must scan its keys, and the
validate entry does neither. Pinned by three tests, including one that strip does NOT delegate,
since a delegated node would silently keep the keys strip exists to drop.

### After

| case | before | after |
|---|---|---|
| parse{preserve} vs `restoreFromJson` + `validate` | 0.64x | **0.98x** |
| parse{preserve} vs validate alone | 0.46x | **0.71x** |
| parse{preserve}, Order | 39.6 M ops/s | **62.8 M ops/s** |

So parse is now at parity with the composition it replaces, where before it was a third slower.

### What is still slower, and why

`parse{strip}` runs at 21.0 M ops/s against validate's 88.7 M. That is the object rebuild: strip
allocates a new object per node to drop undeclared keys, which the composition never does at all.
It is real work for a real guarantee, not overhead, and the comparable reference is
`jsonDecoder{strip}` + `validate`, where the two land within 7% of each other (both dominated by
`JSON.parse`).

For a type that DOES restore, parse keeps its own walk (the validate entry cannot rebuild a Date
from a string) and sits at 0.86x of the composition.

### Method

Pool of 64 distinct values indexed per iteration, results accumulated, 300k iterations, best of 5
rounds. Node v26.8.1, 4-core Intel Xeon @ 2.80GHz. Script not committed.

---

packages/ts-runtypes/src/createRTFunctions.ts
```diff
@@ -521,6 +559,94 @@ export function createJsonDecoderFn<T>(
...
+  // even for recursive schemas), same as createStandardSchema.
+  const runTypeId = isRunTypeValue(valOrSchema) ? valOrSchema.id : undefined;
+  // TWO tuples in Fn-arg order 'prs','verr'. The parse body is the hot path; the
+  // report is only built when something fails, which is why the pair is injected
+  // here rather than composed by the caller.
+  //
+  // `strategy` is compile-time: the plugin already resolved it to one of the
+  // three parse families and baked that family's fnHash into the first tuple, so
+  // the runtime just resolves what it was handed.
+  const injected = ids as unknown as readonly EntryTuple[] | undefined;
```
M-jerez: why do we have to cast to entrytuple here? should InjectTypeFnArgs alreade be typed? seems like a bug to me habing to cast

packages/ts-runtypes/src/runtypes/parseError.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor(value: unknown) {
+    this.value = value;
+  }
+}
+
+/** Error thrown by a `createParseFn<T>()` function. `issues` is the full report,
+ *  identical to `createGetValidationErrorsFn<T>()` over the same value. **/
+export class RTParseError extends Error {
+  readonly issues: RTValidationError[];
```
M-jerez: should this not include serialisation errors as well? `RTValidationError[] | SerializationError` or similar ?

packages/ts-runtypes/src/createRTFunctions.ts
M-jerez: Half right, and the half that is not is worth recording.

The cast cannot go away. `InjectTypeFnArgs` is declared `string & {brand}` on purpose, and the reason is in its doc comment: the Go marker scanner only resolves `T` and the `Fn` keys reliably through the string-intersection alias, and a tuple-intersection alias loses them. So the declared type deliberately does not describe the value it carries at runtime (an array of entry tuples), and something has to bridge that.

What was wrong is that every multi-family factory bridged it itself. `createStandardSchema` had the identical `as unknown as readonly EntryTuple[]`, and the public `jsonValueCodec` example in the docs leans on `fns?.[0]` type-checking only because indexing a string yields a string, which is assignable to `getRTFunction`'s `unknown`. Three call sites, three different ways of papering over the same gap.

Fixed in e576df2: `entryTupleAt(injected, n)` in `entryTuple.ts`, the module that owns the tuple format, so the one unavoidable cast lives in one place and neither factory casts any more.

```ts
const parse = resolveEntryTupleFn<ParseRestoreFn>('createParseFn', parseNoPluginFallback, runTypeId, entryTupleAt(ids, 0));
const getErrors = resolveEntryTupleFn<GetValidationErrorsFn>('createParseFn', getValidationErrorsIdentity, runTypeId, entryTupleAt(ids, 1));
```

---

packages/ts-runtypes/src/runtypes/parseError.ts
M-jerez: You found a real bug. Not the shape you proposed, but the thing you were sensing was there.

I checked rather than reasoned about it, and most of the time the single array is right. A restore arm throws *precisely because* the leaf is still in wire form, so the validator then flags that same leaf at that same path, and the report is exact:

```
{n: bigint} given {n: 'not a number'}  ->  issues=[{expected:'bigint', path:['n']}]
{at: Date}  given {at: 'garbage'}      ->  issues=[{expected:'date',   path:['at']}]
```

A union breaks that, because its members decode through an indexed envelope. The encoder writes `{n:[0,'nope']}` for `{n: bigint | string}`. Hand parse a bare `{n:'nope'}` and the bigint arm throws, but the *undecoded* value still satisfies the `string` member, so validating it comes back clean:

```
{n: bigint | string} given {n: 'nope'}  ->  msg="parse failed"  issues=[]
```

An empty report on a thrown error breaks the one property the design rests on. Fixed in 86b0008, two parts:

- The report falls back to `{path: [], expected: 'jsonForm'}` instead of nothing, with the message reading `parse failed: the value is not the JSON form of this type`. Mapped in `issueMapping.ts` so Standard Schema consumers get a sentence too. Keeping it in `issues` rather than as a second error type is deliberate: a caller already rendering those still needs no second code path, which is the whole selling point.
- The emitted catch now forwards the raw throw as the error's `cause`. That turned out to matter more than I expected. The union restore arm already produces a good message (`Can not json decode union: invalid…`) that the report simply cannot express, and without the cause it was being thrown away.

```js
// before
}catch{ throw utl.parseMismatch(v) }
// after
}catch(e){ throw utl.parseMismatch(v,e) }
```

`cause` is set only on a decode failure. A value that merely fails the check has no underlying throw, and inventing one would say a restore arm broke when nothing did. Pinned by four tests plus a sweep asserting every case in the junk suite reports at least one issue, which is what would have caught this originally.

---

PR #192, opened 2026-08-31
fix(runtypes): hasUnknownKeys runsAfterValidation reported undeclared keys on an array

M-jerez:
Pre-existing bug on `main`, from the spec `docs/todos/huk-fastpath-answers-wrong-for-arrays.md`.

## What was wrong

```ts
type L = {length: number};

createHasUnknownKeysFn<L>()([]); // false, correct
createHasUnknownKeysFn<L>(undefined, {runsAfterValidation: true})([]); // true, wrong
```

An array carries no undeclared object properties: its enumerable keys ARE its elements. Every unknown-keys family answers "nothing undeclared" for one. The fast variant did not, so the two documented ways of writing strict validation disagreed on any shape an array satisfies.

`emitInterfaceHasUnknownKeys` dropped the whole object shape guard under the variant. Dropping the `typeof` / `!== null` half is right, the caller promised the value passed validate. Dropping `!Array.isArray` is not: passing validate does NOT prove the value is not an array, because an array really can satisfy an object shape (`[1, 2]` is a `{length: number}`, `['x']` is a `{0: string}`).

`[]` is the sharp case. The fast path compares enumerable key counts and `length` is not enumerable on an array, so `cntEK([])` was 0 against a declared 1.

## The fix

`arraySkipsHasKeyCheck` (`unknownkeys_shared.go`) gates an object node's PARENT key check on `!Array.isArray(v)`. The chain is `||`, "true = something undeclared is here", so skipping means contributing FALSE:

```go
func arraySkipsHasKeyCheck(v string, check string) string {
	return "(!Array.isArray(" + v + ") && " + check + ")"
}
```

It wraps both of the variant's parent forms, the count compare and the key-array scan fallback. Parent check only: the child descent still runs, guardless as the variant's contract promises, and a validated value carries every declared property whatever its container is. A declared `Item[]` never reaches the gate, so element extras are still reported.

## Tests

- **Go emit** — `TestHasUnknownKeys_RunsAfterValidationSkipsArrayParentCheck` pins the gate on both parent forms and that the `typeof` / null halves stay dropped; `TestHasUnknownKeys_PlainVariantKeepsWholeChainGuard` pins that the plain family gains no second gate.
- **JS** — a new block in `unknownKeys.test.ts` asserts the plain predicate, the fast variant and `createUnknownKeyErrorsFn` all agree on `[]`, `[1, 2]` and `['a', 'b', 'c']` against `{length: number}` and `{0: string}`, plus the declared-array and nested-object-on-an-array cases. Four of them fail without the emit fix.
- **Fuzz** — new **O19 huk-variants-agree** oracle comparing the two variants on every value that passes validate, run in the valid, invalid and junk passes, with an `ArraySatisfiable` target so the junk stream's arrays reach it. 22 violations in 700 runs without the fix, zero with it.

Full JS suite (`pnpm run test:ci`, 7 batches, 10712 tests), `go -C ts-go-runtypes test ./internal/...`, `pnpm run lint` and `pnpm run check-format` all green.

## Docs

The `UnknownKeyErrorsFn` JSDoc had recorded the old behaviour as intentional ("it drops the guards deliberately"); it now says which half goes and which half stays. The validation guide and its compiled example gained a line on the array rule.

## Not covered here

The spec was written against a branch carrying fused `checkUnknowns` validators and their O18 parity oracle. Neither exists on `main`, so the three-way JS comparison uses `createUnknownKeyErrorsFn` as the third family and the fuzz coverage is its own oracle. When the fused families land they carry their own array skip, and O18 and O19 then police the same rule from two directions.

Spec moved to `docs/done/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TCaFALenYJCejjAGrCT6uL)_


M-jerez:
Closing without merging. The fix here is correct and the diagnosis is right, but the array rule is being settled in one place instead of two.

The rule the maintainer set is stronger than this spec assumed: **no family emits an unknown-key check for an array, on any variant.** This PR applies it to `hasUnknownKeys{runsAfterValidation}` only. Meanwhile `claude/ts-runtypes-validation-plan-xehx51` (#187) is changing the same decision for the fused `checkUnknowns` families, and having the two land separately is how they drift apart again — which is exactly the class of bug that started all of this.

So the rule goes in once, covering the blind predicate, the `runsAfterValidation` variant and both fused families together, on the #187 branch. The work in this PR is not lost: the `!Array.isArray` parent gate, the three-way JS agreement block and the `ArraySatisfiable` fuzz target all carry over.

`docs/todos/huk-fastpath-answers-wrong-for-arrays.md` stays open until that lands.

---

PR #265, opened 2026-09-10
Per-route request size limits derived from the types

M-jerez:
Every route resolves its own request limit from its types, the platform adapter knows the limit BEFORE the body is read and reads it the fastest way its runtime offers, an unknown path or batch id is answered without reading the body, and the route lookup happens once per request. Nothing measures a response: what a handler answers is the developer's responsibility. No lint rule ships here.

> The structural Map / Set work this builds on (#268) is merged, and this branch is rebased on `main`. The two delegated findings (the AWS `isBase64Encoded` fix and the unused `getRouteExecutableFromPath`) are merged too.

## What changed

**Request limit per route**, settled once at registration: the route's `maxBodySize` option, else the largest compact JSON its whole chain's params can be (times `maxBodySizeFactor`, a router option, default 2), else the platform adapter's `maxBodySize` (128 KB by default on every adapter). There is no router-wide size any more. A route option wins over everything mion controls and is never clamped by the adapter's number. The name stays `maxBodySize`: it is what Fastify (`bodyLimit`), Bun (`maxRequestBodySize`) and the adapters already call it.

**The adapter resolves the route before the body, and builds the context after it.** A streaming adapter calls `resolveRequest(path, urlQuery, rawRequest)` (the one route lookup, no context yet), reads the body against `resolved.maxBodySize`, then builds the context with the body already in it (`createContextFromResolved(resolved, reqHeaders, respHeaders, rawBody, bodyType)`) and runs the chain with `dispatchWithContext(context, rawRequest, rawResponse)`. The order is load bearing: a context built before the read survives it, is promoted to the old heap, and takes the body assigned into it along, which cost the node lane about 10% throughput and 50 MB at 4 MB bodies. `createCallContext` and `dispatchRoute` keep their signatures for a caller that already has the body. `ResolvedRequest` is `Omit<CallContext, 'request' | 'response' | 'shared'>`, so the two shapes cannot drift.

**The body is read the fastest way each runtime offers.** The shared reader lives in the router (`readRequestBody(req, maxBodySize, strategy)`) with a numeric strategy per runtime, measured with a real server on each: `stream` for a node Request (vercel), where the stream beats `text()` at every size; `text` for workerd (cloudflare), native `text()` with a declared `content-length` and the cancellable stream otherwise; `buffered` for bun, `text()` or `arrayBuffer()` and never the stream reader, which is over ten times slower there. Node keeps its chunk loop (one concat with the running total, the stream destroyed past the limit); uws keeps its native read and decodes the single-read window in place, skipping the decode on a body-less request. AWS Lambda and Google Cloud Functions hand over an already assembled body, so only the router's check before parsing applies there.

**An unknown path or an unknown batch id never reads the body.** Both resolve to a body-free internal chain (`mion@notFound`, the new `mion@batchNotFound`): the adapter skips the read (node answers before the body finishes and lets node discard the rest, uws consumes and drops it, the fetch runtimes leave the stream untouched), the router never parses it, and the global start / end middleFns still run so a rate limiter or an access log sees the request. Route-level middleFns never ran on a 404 and still do not. The batch 404 moves from a bare fatal response to the chain's `@thrownErrors` shape, same status and header.

**JSON size walk** (`ts-go-runtypes/internal/cachegen/jsonsize/`) computes at build time the largest compact-JSON byte size a valid value can have, and whether that is a true bound. The number lands on the trailing slot (`jsonMaxBytes`) of every bounded reflection root; unbounded types carry none. A string bounds by `maxLength`, an array, Set or Map by the `maxItems` of the shared collection params bag. The walk budgets the encoder's wire envelopes, and the `jsonsize` fuzz lane checks it against the compiled encoder's real output on generated bounded types.

**One global switch**: `derivedPayloadLimits` in the `runTypes` block of the mion presets (the resolver's `jsonMaxBytes` option, also a tsconfig plugin key and the `--json-max-bytes` flag, on by default).

**Platform request ceilings**: the four hosted adapters export their documented ceiling in bytes and take it as the `maxBodySizeCap` option; the router never resolves a limit above it.

**Benchmarks**: `bench servers aggregate --compare <before> <after>` prints the change per lane between two results dirs and `sweep <app>` runs the sizes for one lane. Against the branch point, every lane on node, uws and bun sits inside the bench's 10% run-to-run tolerance, so the limits cost nothing on the wire. Against this PR's own first reader (a stream loop with a `TextDecoder` per request) bun's POST lanes run +256% / +178% (about 4,000 to 14,000 and 11,000 req/s). The shipped spec carries the full tables, plus the cloudflare handler under workerd and the reader micro bench.

**Docs**: security page (route lookup first, unknown paths never read the body, how each platform stops a body, the caps, the switch), a new "Unknown Paths" section on the middleFns page, the batch client page, every platform page, the RunTypes configuration page.

## Deferred, each as its own todo
- `docs/todos/lint-platform-payload-caps.md`: build-time diagnostics for an option or a return type above the platform's request or response ceiling.
- `docs/todos/lint-unbounded-types.md`: a diagnostic on any variable-length type without a maximum.
- `docs/todos/failed-before-dispatch-runs-only-always-run.md`: a request refused by the adapter (a 413) still skips the chain, so the always-run middleFns never see it; unifies that with the not-found chains.
- `docs/todos/one-call-context-per-request.md`: reach one context object per request, with the measurements taken here recorded.

https://claude.ai/code/session_011YbsQX1RicX1nFw1VHcdi1


container/website/content/01.rpc/02.server/09.security.md
```diff
@@ -39,19 +39,59 @@ The `x-rpc-error` response header carries the type of the first error. Only a pl
...
+```
+
+Nothing mion resolves passes that ceiling: a route option, a derived limit, a batch sum or a response cap above it is brought down to it, so the metadata never promises a size the platform refuses. The mistake is caught before that, at lint time: the `@mionjs/platform-payload-cap` rule refuses a literal `maxBodySize` or `maxResponseSize` above the ceiling, see [Linter](/rpc/devtools/linter#mionjsplatform-payload-cap).
+
+The resolved number rides the route metadata as `maxBodySize`, so a client can see it. A batch request gets the sum of its routes' limits.
+
 ::tip
-A string body is measured in characters, which is never more than its byte length, so the byte-exact platform limit always fires first where there is one.
+The metadata middleFn sits in every chain and takes a list of ids, so it declares a fixed contribution (4 KB) to every derived limit. A middleFn of your own whose params have no maximum sends every route it guards to the adapter's number; give it `maxBodySize` on its options to declare its own contribution instead.
+::
```
M-jerez: this is internal, not needed

container/website/content/01.rpc/02.server/09.security.md
```diff
@@ -39,19 +39,59 @@ The `x-rpc-error` response header carries the type of the first error. Only a pl
...
+| uWebSockets | The native read stops at the route limit | None | None |
+| Bun | The router checks the route limit before parsing. Bun has one server-wide read limit, which mion sets to the largest limit any route resolves to when the server starts | None (Bun's own default of 128 MB is replaced by that number) | None |
+| Cloudflare Workers | The router checks the route limit before parsing | 100 MB on Free and Pro, 200 MB on Business, 500 MB on Enterprise (the zone plan, not the Workers plan) | None |
+| Vercel | The router checks the route limit before parsing | 4.5 MB | 4.5 MB (a streamed response is not counted) |
+| AWS Lambda | The router checks the route limit before parsing | 6 MB on a synchronous invocation, 1 MB behind an Application Load Balancer or on an asynchronous one; a base64 encoded binary body leaves about 4.5 MB | 6 MB buffered, 200 MB streamed |
+| Google Cloud Functions | The router checks the route limit before parsing | 10 MB (1st gen), 32 MB (2nd gen) | 10 MB (1st gen), 32 MB (2nd gen, 10 MB when streamed) |
```
M-jerez: move this table to a new seciton platfrom cap limits, and remove "## Response Size Limits"

Overall reduce the length of all this documentation , and use super simple language,

container/website/content/01.rpc/06.devtools/01.linter.md
```diff
@@ -189,6 +191,51 @@ type Poison = {
...
+
+The rule needs to know the platform. A file that imports `@mionjs/platform-<name>` names it by itself, so the adapter call is checked with no configuration. Route files do not import the adapter, so give the rule the `platform` option to check them too. A `payloadCap` literal in the same file, or explicit `maxRequestSize` / `maxResponseSize` numbers in the options, replace the table's numbers, which is how a bigger plan (a Cloudflare Business zone, a 2nd gen Google function) or a vendor change is written in. A value that is not a plain number literal is not checked.
+
+| Platform | Request ceiling | Response ceiling |
+| --- | --- | --- |
+| `node`, `uws`, `bun` | None | None |
+| `cloudflare` | 100 MB | None |
+| `vercel` | 4.5 MB | 4.5 MB |
+| `aws` | 6 MB | 6 MB |
+| `gcloud` | 10 MB | 10 MB |
```
M-jerez: how would we handle if there are multiple platforms imports in the code base?
I think this whole feature was adding without much planing, maybe better file a todo for hardening

docs/todos/enrich-check-completeness-gate.md
M-jerez, line 1: this was fixed lets make sure is handled properly

docs/todos/validation-collection-cap.md
M-jerez, line 1: we wont implement this, the arrays and collection already should emmit validation for max lentght et.

please check this is true and add tests fow new Map and Set maxlength

packages/devtools/src/lint/rules/platform-payload-cap.ts
M-jerez, line 1: if I'm not wrong we Already ahave all the lint rules logic in the go compiler, and  are diagnostic errors that are later just proxy to the linter.
let's follow same practice and lets add some CLAUDE.md to the lint directory to make clear the pattern and ensure any future lint rules follow same patter.

We might even need to add a new error diagnostic type, we aready have error, runtimeError and warning, not sure if we would need a new one for the ones that are purely linter?

packages/router/src/routes/client.routes.spec.ts
```diff
@@ -262,7 +267,7 @@ describe('Client Routes should', () => {
       paramsCount: 0,
       paramNames: [],
       pointer: ['last'],
-      options: defaultMiddleFnOpts,
+      options: {...defaultMiddleFnOpts, maxResponseSize: 4},
```
M-jerez: why `maxResponseSize: 4` where is that coming from ?

container/website/content/01.rpc/02.server/09.security.md
```diff
@@ -39,19 +39,49 @@ The `x-rpc-error` response header carries the type of the first error. Only a pl
...
+createCloudflareHandler({maxBodySizeCap: 200_000_000});
+```
+
+Nothing mion resolves passes that ceiling: a route option, a derived limit or a batch sum above it is brought down to it, so the metadata never promises a size the platform refuses. The mistake is caught before that, at lint time: the `@mionjs/platform-max-body-size` rule refuses a literal `maxBodySize` above the ceiling, see [Linter](/rpc/devtools/linter#mionjsplatform-max-body-size).
+
+The resolved number rides the route metadata as `maxBodySize`, so a client can see it. A batch request gets the sum of its routes' limits.
+
+::tip
+The metadata middleFn sits in every chain and takes a list of ids, so it declares a fixed contribution (4 KB) to every derived limit. A middleFn of your own whose params have no maximum sends every route it guards to the adapter's number; give it `maxBodySize` on its options to declare its own contribution instead.
+::
```
M-jerez: not needed, implementation details

container/website/content/01.rpc/02.server/09.security.md
M-jerez, line 1: the whole changes to this section adds to much information, please shor and two the point,
split in two sections

## Body Size Limits
## Platform caps

again reduce cuantity of text and be direct and clearm the platfrom table is ok

container/website/content/01.rpc/05.platforms/01.node-js.md
M-jerez, line 1: please reduce the amount of changes to all these platfrom docs, lest alo not add platfrom caps details here, as those could run stale pretty quick if platfrom make changes.

Actually you add the platform caps either on each platform, or in the secutiry page in the table, buth not both places, maybe each platform make more sense so remove the platform cap details from security and just use generic terms

packages/router/src/router.ts
```diff
@@ -120,14 +123,61 @@ export const getAlwaysAwait = () => alwaysAwait;
...
+  const cap = config?.maxBodySizeCap;
+  return typeof cap === 'number' ? cap : undefined;
+}
+
+/** Nothing mion resolves passes the platform's own request ceiling: a route (or batch) limit, or
+ *  the adapter's number, above it would promise a size the platform refuses before mion runs, so
+ *  it is brought down to the ceiling. Applied once, when the adapter has published its config AND
+ *  the routes are registered, whichever comes last. The `@mionjs/platform-max-body-size` lint rule
+ *  is the build-time half: it flags the option that asked for more. */
+function applyMaxBodySizeCap(): void {
```
M-jerez: move all aucxiliary functions to the bottom of the file.
firt code should be about the router and register routes

packages/router/src/callContext.ts
```diff
@@ -5,29 +5,45 @@
...
+ * request limit: the route's own option, else the number its types derived at registration, else
+ * the platform adapter's `maxBodySize`. The adapter reads the body against `maxBodySize` (node
+ * stops the stream, uws stops the native read) and hands the same handle to `dispatchResolved`, so
+ * nothing is looked up twice. An unknown path resolves to the not-found chain; an unknown batch id
+ * throws a FatalError.
+ */
+export function resolveRequest(path: string, urlQuery: string | undefined, rawRequest: unknown): ResolvedRequest {
+  const opts = getRouterOptions();
+  const transformedPath = opts.pathTransform?.(rawRequest, path) || path;
+  const resolved = getExecutionChain(path, transformedPath, urlQuery, rawRequest, opts) as ResolvedRequest;
```
M-jerez: `as ResolvedRequest` is jus trickery, if we returning the execution chain then use the propername.
saem for resolvedRequest functions name, not what good using a differnt nameand type from the real one could bring us, we also do not want to be mutating execution chain,  this is pure metadata and we should be addin path and query to that object.

So it look sto me that insted adding some trickery, what we need to do from the platfom adapter is just get the context instead the resolved that is mixing some stuff up.

revert any changes to call context, and gte it from the platform adapter, then ww will pass as an optional param to dispatch, what do you think? maybe this require small planning on it's own

ts-go-runtypes/internal/cachegen/jsonsize/jsonsize.go
```diff
@@ -0,0 +1,500 @@
...
+		defer delete(w.inflight, rt.ID)
+	}
+	result := w.walkKind(rt, path, depth)
+	if rt.ID != "" {
+		w.memo[rt.ID] = result
+	}
+	return result
+}
+
+func (w *walker) walkKind(rt *reflection.RunType, path string, depth int) Result {
```
M-jerez, line 150: are you sure we traversing all nodes here?

I think we had somethgin in go that forces code to traverse all nodes, make sure to use it if  we already not doing it

container/website/content/01.rpc/02.server/09.security.md
M-jerez: Done in 6cfde7d: the section is now two short ones, "Body Size Limits" (the three-row order table, the example, one paragraph) and "Platform Caps" (the table, no vendor numbers). The response section is gone.

---

container/website/content/01.rpc/06.devtools/01.linter.md
M-jerez: The hand-written rule is removed from the PR entirely. The check is filed as a guidelines todo (`docs/todos/lint-platform-payload-caps.md`) to be built as a resolver diagnostic, where knowing the platform (several adapters in one code base included) is planned properly.

---

docs/todos/enrich-check-completeness-gate.md
M-jerez: Handled: the fix merged on `main` (with its spec in `docs/done/`), so this branch no longer carries the todo. The PR was restacked on top of that.

---

docs/todos/validation-collection-cap.md
M-jerez: Checked and the todo is deleted. Declared maxima are already refused by the validators: `maxLength` on strings, `maxItems` on arrays, and the new `maxSize` on Map / Set (`v.size <= N`). The tests for Map and Set live in #268: the `sized_map` and `sized_set` entries of the structural suite feed a Map / Set one entry over `maxSize` as invalid samples and pin the `formattedMap` / `formattedSet` `maxSize` error in `expectedFormatErrors`, in both marker shapes. This todo was about something else (capping by the binary sizing knobs), which we will not do.

---

packages/devtools/src/lint/rules/platform-payload-cap.ts
M-jerez: Yes, that is the pattern: the rules are transport for resolver diagnostics. The hand-written rule is gone from this PR, and 6cfde7d adds `packages/devtools/src/lint/CLAUDE.md` stating the rule (emit from the resolver, add a `RULE_SPECS` row, never inspect the AST on your own) and the open severity question. The todo for the platform-cap diagnostics records that a fourth, lint-only level may be needed before the first such diagnostic lands.

---

packages/router/src/routes/client.routes.spec.ts
M-jerez: It was the response cap derived from the `last` middleFn's `boolean` return (`true` is 4 bytes). The whole response side is removed, so this expectation is gone.

---

container/website/content/01.rpc/02.server/09.security.md
M-jerez: Removed both tips in 6cfde7d.

---

container/website/content/01.rpc/02.server/09.security.md
M-jerez: Done in 6cfde7d: "Body Size Limits" and "Platform Caps" are two short sections, the platform table kept, the vendor numbers moved to the platform pages.

---

container/website/content/01.rpc/05.platforms/01.node-js.md
M-jerez: Trimmed in 6cfde7d: each platform page changes one or two sentences, and the ceiling number lives only there (the security page says "Yes" or "None" and points to the page). Node, Bun and uWebSockets name no number at all.

---

packages/router/src/router.ts
M-jerez: Moved in 6cfde7d: the size-limit helpers (`getPlatformMaxBodySize`, `getPlatformRequestCap`, `getMaxRouteBodySize`, `readMaxBodySizeCap`, `applyMaxBodySizeCap`) now sit in a "PLATFORM SIZE LIMITS" block at the bottom of the file, after the private route helpers. `setPlatformConfig` and `getPlatformConfig` stay where they were.

---

ts-go-runtypes/internal/cachegen/jsonsize/jsonsize.go
M-jerez: The helper is `reflection.WalkGraph` (it descends through `EachRefSlot`, every child slot). This walk cannot be built on it on purpose: the JSON size is per kind and reads only the slots that reach the wire (a property's Child, a tuple's Children, a Map's key / value Arguments), and skips Parameters, Return, TypeMeta, the union discriminators and Extends / Implements, which never do. What it needed was the guard, so 6cfde7d adds `TestMaxBytes_VisitsEveryWireSlot`: on a bounded graph spanning every compound kind it sizes, every node `WalkGraph` reaches must have been sized. It caught one gap already (object properties were read around, not through, the property node; fixed). A kind with no arm is never claimed bounded, so a miss can only make a route fall back to the adapter's number, never overstate a limit. The reasoning is now in the package comment.

---

packages/router/src/callContext.ts
M-jerez: Agreed, this needs its own plan. Left open on purpose: the plan (with code examples of how an adapter gets the route data and hands it to dispatch, without the `ResolvedRequest` shape or any mutation of the execution chain) comes next, and nothing in this area changes until it is approved.

---

ts-go-runtypes/internal/cachegen/jsonsize/jsonsize.go
M-jerez: Follow-up in 47100ff, per the chat: the walk is now checked against the real serializer by a fuzz lane (`jsonsize`, `test/fuzz/type/jsonSizeBound`). It generates bounded types (a new typeGen mode: every string a bounded format, every array a `maxItems`, every Map / Set a `maxSize`), compiles them, mocks values, and measures both `JSON.stringify` and the compiled JSON encoder's output against the root's `jsonMaxBytes`; the DataOnly preset runs too. Its first run caught two wire envelopes the walk did not budget: the `[null]` wrap of a root `undefined` / `void`, and the flat union's `[index, value]` wrap. Both are counted now and the lane is green (in `recommended` CI at the quick budget, soak via `MION_FUZZ_JSONSIZE_SOAK_MS`).

---

packages/router/src/callContext.ts
M-jerez: Done in 1b49157, per the plan we agreed: `resolveRequest`, `dispatchResolved` and the `ResolvedRequest` type are gone, the execution chain is never mutated, and `callContext.ts` is back to its shape on `main` plus `maxBodySize`. A streaming adapter now builds the real `CallContext` before the body (`createCallContext(path, urlQuery, rawRequest, reqHeaders, respHeaders)`, the one route lookup), reads the body against `context.maxBodySize`, then calls `dispatchWithContext(context, rawRequest, rawResponse, rawBody, bodyType)`, which attaches the body and runs the chain. `dispatchRoute` keeps its signature (it builds the context with the body and calls the same function), and aws / gcloud, which hold the body already, build the context with it. The dispatch bench is unchanged within noise.

---

PR #266, opened 2026-09-10
feat: bundle the API's route metadata and validators into the client

M-jerez:
## What this does

A mion client used to learn how a route works by asking the server on first use, keeping the answer in the browser and rebuilding the compiled functions with `new Function`. With the new `bundleApi` build option the build does that work instead: for every route the client calls, it compiles the same validators and serializers the server holds and ships them with the call, as real functions. No metadata request, no browser cache, no dynamic code, so a client runs under a strict CSP or on an edge runtime.

- **`bundleApi: 'bundled' | 'mixed'`**, off by default, in all four spellings (resolver option, `--bundle-api`, tsconfig plugin key, Vite and Next presets). `bundled` bundles everything the client calls; `mixed` still fetches a route the build could not see.
- **Markers on the dispatch points.** `.call()`, `.prefill()`, `.typeErrors()` and a batch's `.call()` carry an `InjectApiMetadata<Api, Id>` slot; the route id and the API ride the subrequest type, so a call keeps its route through destructuring and aliasing. `initClient` carries the mode.
- **`apiTsconfig` / `api: {tsConfig}`**, the mirror of `clientTsconfig`: the client build resolves the route types in the API project's own program, so a different `lib` or strictness on the client side cannot change an id.
- **The API type carries each route's effective options** and the types the server compiled it from (`PublicApi`, `ResolvedRouteOptions`, `HandlerMethodTypes`). A route's own `maxBodySize` rides it too; the limit the router settles for a route that declares none exists only at registration, so the type leaves it undefined.
- **Go `apimeta` lane**: walks the API type, selects each called route plus its middleFn chain, assigns ids under the owning checker, emits one live-function module per method under `<genDir>/api/` (built-in pure fns included), and injects the site module at the call. Diagnostics MET001 to MET006, in their own section of the diagnostics catalog.
- **Manifests and `mion api-check`**: both builds write `<genDir>/api/manifest.json`; `mion api-check --server-gen-dir --client-gen-dir` compares them and exits 0 / 1 / 2, a prerelease gate for a split deployment.
- **Client runtime**: bundled payloads register through the same reflection the router runs, hash for hash, and a bundled method carries exactly what a fetched one does. `BundledMethod` is a `Pick` of core's own `MethodWithOptions`, so the two cannot drift. The server's params byte ceiling stays on the server, since it only settles a chain's request limit at registration. `bundled` never fetches, `mixed` fetches only what the bundle lacks. A payload the build did not write is reported in the result's undeclared slot, because a call never throws.

## Related fixes found on the way, each in its own commit

- `getSerializableMethod` never copied `headersReturn`, so a route returning headers never gave them back.
- Optional parameter names reached the metadata wire union-tagged instead of as plain strings, and an unlabelled one reached it as null.
- An entry tuple was never registered again after its cache entry was removed.
- Built-in pure fns in the bundled mirror were rendered in the session's emit mode, so a `code`-mode program shipped them as code strings.
- The six new diagnostics mapped to no catalog section, so they rendered on no page. An unmapped code family now fails codegen instead of warning.

## Tests

- Go: `apimeta` and `resolver/apigen` suites (discovery, tree walk, chains, every diagnostic, generation, both wire modes, the `apiTsconfig` peer program, manifests across two projects), including the paired `getRunTypeId` test the marker coverage rule asks for, and `cmd/mion` api-check.
- Client: two new vitest lanes (`client-bundled`, `client-mixed`), each its own program with its own test-server process, plus a parity test against the live server's metadata answer.
- Devtools: a real vite build of a bundled client (live factories, no code string, runs with `Function` disabled), `mion compile --bundle-api --api-tsconfig` then `mion api-check` through the three exit codes.
- Fuzz: the new `apiids` lane, the real binary over a server project and a client project with a different tsconfig per generated type (`pnpm miondevx core fuzz apiids --quick`).

## Type budgets

Four budgets were raised, which the budget rule allows only as a reviewed exception, so calling them out: steps 4 and 5 of the model pipeline and two alternative-lane numbers. Each records the exact measured cost of what the API type now carries per method, after keying the client and public API mappings on the type discriminant recovered about 1300 instantiations. The pipeline total keeps its existing slack.

## Two exported types changed meaning

`NonClientRoute` and `NonClientMiddleFn` were unions of the public method types and are now structural leaf types, a consequence of `ClientRoutes` and `ClientMiddleFns` dispatching on the `type` discriminant. Nothing in the repo names either, but a consumer that does will see a different shape.

## Docs

New page `rpc/client/bundled-api`, plus the security, Vite, Next.js, CLI and runtypes configuration pages; examples under `packages/examples/src/client/` and `src/codegen/`. The spec moved to `docs/done/` describing what shipped.

https://claude.ai/code/session_01MG6jJr6SJoLANLvX7TtLag


M-jerez:
## `pre-publish e2e · consumer lanes` is failing, and it is not this PR's

Adding the `pre-publish-e2e` label ran that lane on this branch for the first time, and it fails in the CLI compile step:

```
/e2e-mion/src/tests/json.spec.ts(133,82): error BAT008(/e2e-mion/client-app/tsconfig.json)
e2e: the mion consumer lane failed
```

### Why it is not this PR's

The consumer fixture compiles with `--client-tsconfig client-app/tsconfig.json`, and its own program contains a `batch()` call (`container/pre-publish-e2e/mion-consumer/src/tests/json.spec.ts:133`, reached because `tsconfig.compile.json` extends the root tsconfig and includes `src/tests/`). BAT008 is exactly that situation: a batch written in the server program never reaches the table generated from the client project.

BAT008 became fatal on `main` in `ed06912` (`fix(diagnostics): split Error into fatal Error and RuntimeError`), which set it to `LevelRuntimeError`, and a build lane exits non-zero on that level. Every input to the failure is untouched by this branch:

| Input | Where it lives | Changed here |
| --- | --- | --- |
| The BAT008 level | `internal/diagnostics/codes_batch.go` | no |
| The condition that emits it | `internal/compiler/resolver/rpcgen.go`, `collectRpc` | no |
| The fixture with the batch | `container/pre-publish-e2e/mion-consumer/` | no |
| The compile command | `mion-consumer/package.json` | no |

This branch's `rpcgen.go` edit only lifts the peer-session fields into a shared `peerProgram` so the new `apiTsconfig` lane can reuse them; the `separate := source != sess` test and the diagnostic it guards are unchanged.

The lane is label gated, so nothing ran it between `ed06912` and today, which is why it went unnoticed.

### What happens next

No fix for it exists yet, so there is nothing to port into this branch. I am opening it as its own change against `main`; once that merges I will bring it in here and re-run. The proposed patch is to stop the CLI compile fixture from pulling the vitest specs into the server program, so its `batch()` lives only in the client project the pointer names:

```jsonc
// container/pre-publish-e2e/mion-consumer/tsconfig.compile.json
{
  "extends": "./tsconfig.json",
  "exclude": ["src/tests/**", "client-app/**", "dist", "dist-cli", ".mion", ".mion-cli"]
}
```

I have not re-run the lane, since a re-run reproduces the same deterministic diagnostic rather than telling us anything new.

---

M-jerez:
Following up on the comment above: the `pre-publish e2e · consumer lanes` failure is being fixed in its own change against `main`, not here, since it is unrelated to this branch. The spec is `docs/todos/pre-publish-e2e-batch-in-the-server-program.md` (added here as the handoff point, and it moves to `docs/done/` in that PR).

**Merge order:** that fix merges first. This PR waits for it, then rebases on `main` and drops its copy of the spec.

---

PR #268, opened 2026-09-11
Structural Map and Set formats on the array keywords

M-jerez:
Nothing bounded a Map or a Set until now. This adds the missing structural format family, planned and built from the spec in `docs/done/structural-map-and-set-formats.md` (the branch was rewritten from that spec; the first version of this PR was a one-shot draft).

## What changed

**Both collections reuse the ARRAY keywords.** A Set is an array on the wire and a Map is an array of `[key, value]` pairs, so the bounds land in the generated JSON Schema under the standard names, the size estimator reads the same `maxItems` it reads for arrays, and one Go helper serves the three families.

- `TF.FormattedSet<Set<T>, P>` takes `FormattedArrayParams` whole: `minItems` / `maxItems` (read off `.size`), `uniqueItems` (deep JSON equality over the members, meaningful for object members) and `contains` / `minContains` / `maxContains`.
- `TF.FormattedMap<Map<K, V>, P>` takes `FormattedMapParams`, a `Pick` of the array bag's two count keys.
- `RT.set(item, {...})` and `RT.map(key, value, {...})` are the value-first twins, one id with the type-first spelling. Same overload shape as `array` / `record` / `object`: a single signature with an optional bag was built and measured against the type-cost budgets and rejected (+7 first call per builder, +4 per call with a bag, about +95 per utility wrapper, +111 on nested arrays); the numbers are recorded in `builderCost.compile.test.ts` next to the other rejected experiments.

**Go side.** `collectionformat.go` registers `formattedSet` / `formattedMap` under `KindClass` and shares the count-bound helpers with `arrayformat.go`. Both intersection collapses (serialize and typeid) now promote a `contains` slot onto a Set base, which they silently dropped before, and the `contains` count in the validate and errors lanes walks a Set with `for…of` (arrays keep their index loop byte for byte). The JSON Schema output carries the bounds on the outer array, the binary size estimate reads `maxItems` on Map / Set, and `mion convert` prints both families in both directions.

**JS side.** The `rt::uniqueItems` pure fn walks a Set (only object members are canonicalised), the Set mock satisfies `contains` and `uniqueItems` by construction, `StripRunTypeMeta` rebuilds a branded collection, and `DataOnly` already keeps a branded collection whole (pinned by a compile test).

**A fix found on the way:** the errors lane guarded every class format check with `v instanceof Date`, so no Map / Set check ever ran there. First commit.

**Tests.** Go: emitter, module render, collapse, typeid, schema render, size estimate, convert roundtrip, catalog sync. JS: four cases in the structural suite through the 12-lane matrix, three value-first cases, the bare-id and twin-id pins, the `dataonly` / `stripmeta` compile tests, and the convert fuzz lane now draws bounded Set / Map shapes.

**Docs.** The type-formats page gains the Set / Map wrappers and rows, with a `<code-import>` example; the JSON Schema page notes where the bounds land.

**Not included:** the `List` / `SizedMap` / `SizedSet` sugar aliases (dropped by decision, the array/object family ships none).

PR #265 is stacked on this branch and reads `maxSize` in its `jsonsize` walk; it needs a rebase and a switch to `maxItems`.

https://claude.ai/code/session_01Ktxj3mEUZ9wwQbEiS11T54


packages/run-types/src/builders/compose.ts
```diff
@@ -338,19 +342,48 @@ export function record(
...
+export function map(
+  keySchema: RunType,
+  valueSchema: RunType,
+  arg3?: FormattedMapParamsValueFirst | InjectRunTypeId<unknown>,
+  arg4?: InjectRunTypeId<unknown>
+): RunType {
+  const base = {type: 'map', index: keySchema, child: valueSchema};
+  if (isFormatParams(arg3)) return builderResult(arg4, base);
+  return builderResult(arg3 as InjectRunTypeId<unknown> | undefined, base);
 }
```
M-jerez, line 375: not sure why we need a function overload here, does just optional params do not work ?

packages/run-types/src/builders/compose.ts
```diff
@@ -338,19 +342,48 @@ export function record(
...
+): RunType<FormattedSetFrom<Set<V>, P>>;
+export function set(
+  valueSchema: RunType,
+  arg2?: FormattedSetParamsValueFirst | InjectRunTypeId<unknown>,
+  arg3?: InjectRunTypeId<unknown>
+): RunType {
+  const base = {type: 'set', child: valueSchema};
+  if (isFormatParams(arg2)) return builderResult(arg3, base);
+  return builderResult(arg2 as InjectRunTypeId<unknown> | undefined, base);
 }
```
M-jerez: samethign here, why not optional params, no overload?

packages/run-types/src/formats/structural.ts
```diff
@@ -159,6 +176,54 @@ export type FormattedObject<Base extends object, P extends FormattedObjectParams
...
+export interface FormattedMapParams {
+  readonly minSize?: number;
+  readonly maxSize?: number;
+}
+
+/** The entry-count bounds of a Set: `minSize` / `maxSize`, read off `.size`. **/
+export interface FormattedSetParams {
+  readonly minSize?: number;
+  readonly maxSize?: number;
+}
```
M-jerez: should we add similar options like FormattedArrayParams or FormattedObjectParams

packages/run-types/src/builders/compose.ts
M-jerez: Optional params do work mechanically (the transform pads a skipped optional bag with `undefined`, and both marker shapes resolve), so it was built for all five builders and measured against `builderCost.compile.test.ts`. It loses on type cost: +7 first-call instantiations per builder even with the result type gated on the empty bag, +4 per call carrying a bag, about +95 on each utility wrapper (`partial(object)` 905 to 998) and +111 on the nested-array base (373 to 484). Without the gate the first call is 4 to 5 times the budget. Those budgets are a one-way ratchet, so the overloads stay, same as `array` / `record` / `object`; the measurement is recorded in the budget file's rejected list so nobody re-runs it.

---

packages/run-types/src/builders/compose.ts
M-jerez: Same answer as the `map` thread: the single-signature shape was built and measured, and the overload is what keeps a bare `set(string())` at its plain type cost (72 first-call instantiations instead of 79 or more). Details and numbers in the other thread and in `builderCost.compile.test.ts`.

---

packages/run-types/src/formats/structural.ts
M-jerez: Done, and further than "similar": a Set now takes `FormattedArrayParams` itself (`minItems`, `maxItems`, `uniqueItems`, `contains`, `minContains`, `maxContains`), and `FormattedMapParams` is a `Pick` of that bag's two count keys, so the vocabulary cannot drift. A Set is an array on the wire and a Map an array of pairs, which is why the array keywords fit: the generated JSON Schema carries them under the standard names and the size estimator reads the same `maxItems`. The `minSize` / `maxSize` names and the separate interfaces are gone.

---

PR #272, opened 2026-09-12
fix(e2e): keep batches out of the mion consumer's --client-tsconfig program

M-jerez:
## What was broken

The `pre-publish e2e · consumer lanes` job failed on `main`:

```
/e2e-mion/src/tests/json.spec.ts(133,82): error BAT008(/e2e-mion/client-app/tsconfig.json)
e2e: the mion consumer lane failed
```

The consumer fixture compiles its server with a separate client project:

```
mion compile --tsconfig tsconfig.compile.json --client-tsconfig client-app/tsconfig.json --gen-dir .mion-cli
```

`tsconfig.compile.json` extended the fixture's root tsconfig, so the server program pulled in `src/tests/`, and `json.spec.ts` calls `batch([...])` twice. With a client pointer set the batch table is built from that client project alone, so a batch written in the server's own program never reaches it. That is BAT008, and it was promoted from a warning to `LevelRuntimeError`, which exits the build non-zero. The engine is right; the fixture was wrong.

The lane is label gated, so nothing ran it between the promotion and now.

## The fix

`src/tests/` leaves the CLI compile program:

```jsonc
// container/pre-publish-e2e/mion-consumer/tsconfig.compile.json
"exclude": ["node_modules", "dist", "dist-cli", "src/tests"]
```

The specs' batches stay where they are. Both other lanes (the vitest round trip and `vite build`) compile against the fixture's root `tsconfig.json` with no client pointer, and `build-output.spec.ts` asserts that `vite build` inlines the mapper body authored in `json.spec.ts`. Moving the batch into `client-app/` would have traded one broken lane for another, so only the CLI compile changes. The fixture still proves what it proved: a consumer compiling a server against a separate client project, then compiling that client, batches included.

## Guards, since nothing on the host runs the fixture

- `packages/devtools/test/compile-cli-mion.test.ts` gains a case: two real projects plus one `batch()` in the server's own program, asserting `mion compile` exits non-zero and names BAT008 and the offending file.
- `packages/devtools/test/cli-surface.test.ts` gains the fixture guard, in the spirit of the verb/flag guard already there: read the fixture's own `compile` argv, resolve the include and exclude globs of the tsconfig it names, and fail if any file of that program calls `batch(`. A third case asserts the excluded specs do batch, so the guard cannot pass hollow.

Both run in the default test lane, no container and no label.

## Docs

The promotion left two pages reading as if BAT008 were a warning:

- `01.rpc/06.devtools/02.vite.md` said the build "says so". It now says the build fails and how to fix it.
- `01.rpc/03.client/03.batch.md` never mentioned the case. It gains a short note.

## Also carried

One commit ports a one-line fix already on another branch: `.claude/skills/review-pr/SKILL.md` named a spec path, which the tracked-tree contract test forbids. That test is red on `main` today, so this PR could not be green without it.

## Verification

- `pnpm miondevx release e2e --backend container --no-matrix`: `pre-publish e2e: PASS` locally, and `pre-publish e2e · consumer lanes` is green on this PR
- `pnpm run lint` (both linters plus typecheck): green
- `pnpm run test:ci` (all 21 vitest projects): green
- `pnpm miondevx website check`: `smoke: PASS` on this branch

## Merge order

A delegated PR fixing a pre-existing `jsonMaxBytes` undercount (the `jsonsize` fuzz soak, tracked in `docs/todos/json-size-bound-underestimates-a-bounded-set.md`) **merges before this one**. That bug is what reddens `js tests + lint` here; the [comment below](https://github.com/MionKit/mion/pull/272#issuecomment-5648896335) has the evidence that it is not this PR's.

This PR is itself expected to merge before #266.

## Labels

`pre-publish-e2e` (the lane this fixes) and `website` (the docs tree changed).

https://claude.ai/code/session_0148Aqas9a3atGQtHTVLH4nA


M-jerez:
### The lane this PR fixes is green

`pre-publish e2e · consumer lanes` **passed**, along with `pre-publish e2e · build + pack`, `build the docs site`, `go tests + fuzz`, `commit messages` and `detect changes`.

Two checks are red. Neither is this PR's, and here is the evidence for each.

### `js tests + lint` — a pre-existing `jsonMaxBytes` undercount, delegated

The failure is the `jsonsize` fuzz soak, in the `Time-boxed fuzz lanes at quick budgets` step:

```
FAIL  test/fuzz/type/jsonSizeBound.integration.test.ts
AssertionError: expected [ …(3) ] to have a length of +0 but got 3

[jsonsize-fuzz][JS-MAX-ENCODER] [1d] Set<({2}|{3}|{3}|{1})>≤1 (seed=2646330014):
the compiled JSON encoder emits 70 bytes, over the type's jsonMaxBytes 68
    [[-1,{"kind":"t2","f0":[2,null],"f1":[1,"1995-03-12T13:06:34.149Z"]}]]
```

This PR changes a container fixture's `tsconfig.compile.json`, two `packages/devtools/test/` files, two website markdown pages and one skill markdown page. It touches no part of the bound computation or the JSON encoder, so it cannot cause this.

It is a real bug, not a flake: the bound is genuinely below what the encoder emits for a bounded `Set` of a union. It surfaced here because that soak is time-boxed (`MION_FUZZ_JSONSIZE_SOAK_MS`), so each run walks however many types fit the budget and this runner happened to reach that shape. For the same reason the lane's own seed does not replay elsewhere: `MION_FUZZ_SEED=0xe07d86b5 pnpm miondevx core fuzz jsonsize --quick` passes on a dev box. The per-type seed in the report is the replayable handle.

No fix for it exists yet, so there is nothing to port into this PR, and writing one here would widen a fixture fix into the size-bound engine. Per the findings rule in `CLAUDE.md` it is delegated to a parallel session with its own branch and its own PR, carrying the type, the seed, and the reason the outer seed does not replay. **That PR is expected to merge before this one.**

### `container smoke` — the docs dev server died starting up

```
==> smoke: server answered HTTP 500, waiting for 200
==> smoke: FAIL (no 200 from the in-container probe within 90s; last status 500)

ERROR  [request error] [unhandled] [GET] http://127.0.0.1:3000/
Error: IPC connection closed
  at Socket.onClose (node_modules/.../@nuxt/vite-builder/dist/vite-node.mjs:140:101)
```

That is the Nuxt vite-node worker dying during startup, not a content parse error. Two things say the content is fine:

- `pnpm miondevx website check`, the exact command this step runs, passes on this branch with the same image: `==> smoke: PASS  <title>TypeScript APIs and runtime types - mion</title>`
- `build the docs site` on this same head is green, and that is a full static build of the same content tree including both pages this PR edits.

Re-running both jobs once.

---

PR #276, opened 2026-09-13
fix(router): a request that arrived failed runs only the alwaysRun middleFns

M-jerez:
Implements `docs/todos/failed-before-dispatch-runs-only-always-run.md`, moved to `docs/done/`.

Rebased on `main` now that #281 (the delegated gcloud fix) has merged.

## The problem

Two gaps with the same shape.

**A request the adapter refused never reached the router.** `fatalFail()` wrote a bare response, so no chain ran at all: a rate limiter or an access log registered with `addStartMiddleFns` / `addEndMiddleFns` never saw a 413. That also disagreed with the router's own limit, which throws inside `deserializeRequestBody`, an `alwaysRun` member INSIDE the chain, so aws and a direct `dispatchRoute` already behaved correctly.

**An unknown path ran every global start middleFn**, because the error only appeared when the not-found route threw halfway down the chain. A session loader or an authorization step did real work for a request already known to fail.

## The mechanism

No new state, and **`runExecutionChain` gained no code at all**: no extra read, no extra branch, so the happy path pays nothing. The loop already had the rule:

```ts
if (response.hasErrors && !executable.alwaysRun) continue;
```

so the whole job was making the response carry the error BEFORE the first global middleFn.

| Source | Key | How it fails first |
| --- | --- | --- |
| unknown path | `mion@notFound` | the first member of its chain throws |
| unknown batch id | `mion@batchNotFound` | the first member of its chain throws |
| adapter refusal after the route resolved | `mion@platformError` | the adapter calls `dispatchPlatformError` |

### The two not-found routes became chain members

They are raw middleFns now, throwing the same `FatalError` under the same ids, placed at the head of their own chain:

```ts
const methods = [getExecutableFromRawMiddleFn(middleFnDef, [id], 0), ...startMiddleFns, ...endMiddleFns];
notFoundChains.set(id, {routeIndex: -1, methods, serializer: getChainFraming(methods), readsBody: false});
```

Raw middleFns rather than routes: nobody declared the request, so there is no params or return contract to compile and the two `DEFAULT_WIRE` encoder pins go with them. The throw is keyed by the member's own id, so the wire shape is what it was. `/mion@notFound` and `/mion@batchNotFound` stop being reachable paths and now 404 like any other unknown path. `notFoundChains` goes through `getOrCreateGlobal` like every other router map, and `index.ts` names `mionErrorsRoutes` explicitly so neither thrower joins the published surface.

### The adapter refusal

`recordUndeclaredError(context, key, err)` replaced `onExecutableError`: same `markFatal`, `markResponseFailed` and `thrownErrors` slot, keyed by the caller rather than by a chain member. One new export sits beside the loop, never inside it:

```ts
export function dispatchPlatformError<Req, Resp>(resolved, platformError, reqHeaders, respHeaders, rawRequest, rawResponse?) {
  const context = createContextFromResolved(resolved, reqHeaders, respHeaders); // no body
  recordUndeclaredError(context, MION_ROUTES.platformError, platformError);
  return dispatchWithContext(context, rawRequest, rawResponse);
}
```

Which call sites switched was decided by one question, had the route already resolved? Node's declared `content-length` and mid-read checks, uws' `collectBody` null answer, the `readRequestBody` throw in bun, cloudflare and vercel, and gcloud's own size check (see below). A throwing `pathTransform`, a connection error, a failure while replying and the post-dispatch catch-all keep `fatalFail`, since no chain exists for them. aws is untouched: it hands the body over whole, so its 413 comes from inside the chain.

`deserializeRequestBody` also returns early when `context.response.hasErrors`, so a body a caller handed over anyway is never parsed. That member declares `alwaysRun`, so the dispatcher never skips it and the guard has to be its own.

`toRpcError` moved into the same module: node and uws each carried a private copy and the new call sites needed one, so five identical copies became one export.

### gcloud, after the rebase

#281 gave gcloud its own size check, because express may hand the adapter a body it already parsed and the router only measures a string. That check threw past the chain into a bare answer, which is exactly the shape this PR removes everywhere else, so it now routes through `dispatchPlatformError` too:

```ts
try {
  rejectOversizedRequest(rawRequest, rawBody, resolved.maxBodySize);
} catch (refusal) {
  const refused = await dispatchPlatformError(resolved, toRpcError(refusal), reqHeaders, respHeaders, rawRequest, rawResponse);
  return reply(refused, rawResponse);
}
```

The gcloud test that catches it was already on this branch and failed on the merged tree (`expected [] to deeply equal [ 'log:413' ]`, so not even the `alwaysRun` member ran). One sentence in #281's `docs/done` entry, which said those middleFns do not run, is corrected with it.

### What each adapter does with the refused body

Measured with the slow-upload technique from #277 (raw socket, 3 MB announced, trickled at 200 KB/s past uWS' 10s idle timeout) against a 64-byte limit:

| adapter | bytes of the over-limit body accepted | socket after | next request on it |
| --- | --- | --- | --- |
| node | 0 / 3,000,000 (`httpReq.destroy()`) | destroyed | n/a |
| uws | 3,000,000 / 3,000,000 (`collectBody` keeps draining) | alive | 200 |
| bun | unread chunks remain, so it sends `connection: close` | closed | n/a |

## The wire shape did not change

The 413 envelope is byte-identical: the exact-wire pins in `mionHttp.spec.ts` and `uwsHttp.spec.ts`, literal `content-length: '152'` included, pass untouched.

## Tests

- `notFound.spec.ts`, rewritten: for each of the three sources, a global `alwaysRun` start AND end middleFn runs, a global one WITHOUT it does not, the route and route-level middleFns do not, `'{not json'` is never parsed, and the answer carries the right key, status and header. Plus the parse guard above, which nothing pinned.
- Every adapter: a body over the route's limit answers 413, an `alwaysRun` end middleFn sees it, a plain global does not, and the server still serves the next request. Node covers the declared `content-length` and the mid-read case. Bun's suite declares a second route with a large `maxBodySize` to lift `Bun.serve`'s own ceiling clear of the test body, without which Bun refuses natively and mion is never called.
- aws pins that it already ran the chain, and its 413 is raised from INSIDE the chain (`mionDeserializeRequest`), so the plain global runs first: that is the real difference between an adapter refusal and a router one. gcloud, which had no 404 coverage at all, pins the adapter-refusal side of the same rule.
- `maxBodySize.spec.ts` pins the invariant behind the gcloud hole: the router refuses an over-limit STRING body and deliberately lets the same payload through as an object its host already parsed, which carries no wire size here. That is why an adapter handed a parsed body owns the check.
- `bodyDrain.spec.ts` (from #277) gains a third slow upload, at `/api/echo`: a body refused mid-flight is answered before it finishes arriving, and every announced byte is still accepted.

Full suite green (483 test files in one vitest process), plus `test:bun`, lint and typecheck.

## Behaviour change

A global start middleFn without `alwaysRun` no longer runs on a 404. That is the point, and it is unreleased, so no consumer depends on it. The docs and the example move with it: the middleFns page's "Unknown Paths" became "Requests That Already Failed" covering all three, the security and batch pages drop the old rule, and the example's rate limiter declares `alwaysRun`. The pages also say that a 413 goes through your middleFns either way: when the adapter refused the body only the `alwaysRun` ones run, and on AWS Lambda, where the router refuses it instead, the middleFns registered before the body parse run first.

https://claude.ai/code/session_01H667UCpi4WyyYcPg57CUeF


packages/router/src/lib/dispatchError.ts
```diff
@@ -75,9 +75,28 @@ export function onExecutableError(context: CallContext, executable: RemoteMethod
...
+  recordArrivalError(context, path, rpcError);
+}
+
+/**
+ * Records an undeclared error under `key` and ends the request: the same slot, status and header a
+ * thrown one takes, keyed by the caller rather than by a chain member. What an adapter reaches for
+ * when it refuses a request whose route already resolved (a body past the limit), so the chain still
+ * runs its `alwaysRun` members over the refusal instead of being skipped altogether.
+ */
+export function recordArrivalError(context: CallContext, key: string, err: any) {
```
M-jerez: this name is misleading, it is applied to any error, not only arrival, so use a generic name

packages/platform-bun/src/bunHttp.ts
```diff
@@ -69,7 +70,17 @@ export async function bunRequestHandler(req: Request): Promise<Response> {
...
     // a not-found chain (an unknown path or batch id) has no route to feed: its body is never read
     if (resolved.readsBody) {
-      rawBody = await readRequestBody(req, resolved.maxBodySize, BodyReadStrategy.buffered);
+      try {
+        rawBody = await readRequestBody(req, resolved.maxBodySize, BodyReadStrategy.buffered);
+      } catch (e) {
+        const refusal = toRpcError(e);
+        // a body refused mid-flight leaves unread chunks on the socket: close it with the answer so
+        // they are never parsed as the next request of a kept-alive connection
+        if (refusal.type === 'request-payload-too-large') responseHeaders.set('connection', 'close');
```
M-jerez, line 80: shouldn't we be applyning same connection closed to other servers that might ignore the rest of the connection?
specially node and uws

packages/router/src/lib/dispatchError.ts
M-jerez: Agreed, renamed to `recordUndeclaredError` in c964c54.

It also made a duplication obvious: `onExecutableError` was building the very same `RpcError` before handing it over, so the fallback construction ran twice. It now just passes the member id:

```ts
export function onExecutableError(context: CallContext, executable: RemoteMethod, err: any) {
  recordUndeclaredError(context, executable.id, err);
}
```

and the doc comment says what the key is for rather than describing one caller:

```
 * Records an undeclared error under `key` and ends the request: it lands in `@thrownErrors` untyped,
 * sets the status and the error header, and stops every later chain member that does not declare
 * `alwaysRun`. The key is normally the chain member that failed; an adapter passes its own
 * (`mion@platformError`) for a request it refused before any member could run.
```

---

packages/platform-bun/src/bunHttp.ts
M-jerez: Good question. I measured both rather than reasoning about it, using the slow-upload technique #277 established (raw socket, 3 MB announced, trickled at 200 KB/s so the upload outlives uWS' 10s idle timeout), against a 64-byte limit:

| adapter | 413 answered | bytes of the over-limit body accepted | socket after | next request on the same socket |
| --- | --- | --- | --- | --- |
| node | yes | 0 / 3,000,000 | destroyed | n/a, gone |
| uws | yes | 3,000,000 / 3,000,000 | alive | 200 |

Neither wants `connection: close`, for opposite reasons:

**node** already does something stronger. `dispatchRefusal` ends with `httpReq.destroy()`, so the socket is gone rather than merely marked non-reusable, and it accepts none of the oversized body at all. `connection: close` on top would change nothing.

**uws** drains it properly and the connection stays good. `collectBody` keeps its own `onDataV2` registered after it answers null, so the rest of the body is read and dropped, and (per #277) that same callback is what refreshes the idle timeout. Sending `connection: close` there would throw away a working connection for no benefit.

**bun** is the one that needs it: it can neither drain nor destroy, since its `Request` hands out no socket handle, so the unread chunks would be parsed as the next request on a kept-alive connection. Cloudflare and Vercel need nothing either, their `readRequestBody` cancels the stream mid-flight and the runtime owns the socket.

I pinned the uws half in CI rather than leaving it as something I checked once: `bodyDrain.spec.ts` now trickles a third upload at `/api/echo` alongside the two it already had, asserting 413, all 3 MB accepted and no hang-up. Leaving this thread open since no code changed here, so you can close it if you would still rather have the header everywhere.

---

PR #278, opened 2026-09-13
refactor(router): the unknown-key pair follows the wire, like every other slot

M-jerez:
Rebuilt on the current `main`. The codec half this branch used to carry is already there and better done, so this is now only the slot change.

Implements the open spec for the unknown-key slots, moved to `docs/done/` and rewritten to match what shipped.

## The problem

Every slot a route asks the build for is computed from its encoder strategy, so a route compiles only what its wire needs. `huk` (hasUnknownKeys) and `uke` (unknownKeyErrors), the pair behind `strictTypes`, were the two exceptions: fixed strings, compiled by every route in every app, on both directions, and shipped to the browser by `serializeMethodDeps`.

## The rule

```ts
// packages/router/src/types/encoder.ts
type UnknownKeys<Strategy, Key> = Strategy extends 'clone' | 'compact' ? never : Key;
```

It mirrors `DecodeFamily` directly above it. `clone` decodes with `rjs` and `compact` with `cjr`, and both rebuild the params from the declared type, so a property the type does not name is gone before validation and before the handler. `mutate` and `direct` decode with `rj`, which restores in place and keeps every key, so those two keep the pair.

`clone` is the default params encoder, so most routes now compile neither function.

`returnFns` drops both outright. Nothing reads `returnJitFns.hasUnknownKeys` on any wire: the answer side is written by the handler, never by a caller.

`packages/core/src/runtypes/mionAdapter.ts` leaves the pair off the built set when the marker did not ask for it, instead of standing in a function that always answers false, so both readers take their existing `!hasUnknownKeys` early return.

## One behaviour change to know about

The test server's `createUserStrict` set `strictTypes: true` on the default wire, where it no longer has anything to check. It moved to the wire where it means something:

```ts
createUserStrict: route((_ctx, user: User): User => user, {strictTypes: true, encoder: {params: 'mutate'}}),
```

That is the shape any app in the same position needs. It matches what `dispatch.spec.ts` already does on `main`.

## Tests

- `encoder.spec.ts`: the pair per wire, including the default one, that only the params direction decides, that a middleFn follows its router-wide wire, and that no wire compiles it for the answer side.
- `mionAdapter.spec.ts`: a marker payload with no `huk`/`uke` builds a set without them, untested before.
- `validation.spec.ts` and the test server: the client's local pre-validation still gates on a wire that keeps the pair.

## Not in scope

**The lint rule the spec also asked for was built and deliberately withdrawn.** It would report on `encoder`, an option that names one strategy but silently decides four things (encode and decode, on each direction), with four names that describe only the encoding half. Naming that condition in a diagnostic message and a lint rule name would pin the confusing shape. That option's redesign is filed as its own spec, and the moot-`strictTypes` signal is written into its scope rather than left loose.

The `fmt` slot (the `sanitizeParams` lane) is still a fixed string. It follows an option, not the wire.

## Docs

None. The website's validation page on `main` already states the behaviour this follows: a `clone` or `compact` route drops an undeclared property before validation, so `strictTypes` belongs with `mutate` or `direct`. Nothing a reader sees changed.

## Checks

`go -C ts-go-runtypes test ./internal/... ./cmd/...` (no Go files changed), the router, core, client and test-server suites, `pnpm miondevx core codegen all --check`, `pnpm run format`.

https://claude.ai/code/session_01SJ19UUhNRGbmA5PFZSseEo


container/website/content/01.rpc/02.server/07.validation.md
```diff
@@ -66,6 +66,10 @@ You can enable `strictTypes` globally via router options or per-route:
 When a type has an index signature (e.g. `[key: string]: any`), `strictTypes` is automatically skipped since the type explicitly allows arbitrary properties.
 ::
 
+::note
+The `compact` wire has no unknown keys to find. It sends an object as an array of its values, so the names a caller writes never travel, and a route on that wire compiles no unknown-key check at all. Setting `strictTypes` there is a linter error (`@mionjs/unreachable-strict-types`), since the option would promise something the route cannot do. Turning `strictTypes` on for a whole router is fine: it applies to the routes that can use it.
+::
+
```
M-jerez: note not required, revet changes here

PR #280, opened 2026-09-14
fix(runtypes): every family that strips undeclared keys answers the same

M-jerez:
## What

Seven public functions answer one question: which keys on this value are not declared by the type? Five of them disagreed, each at a different position. Everything here faces UNTRUSTED input, and `validate` accepts undeclared keys on an object literal by design, so nothing downstream caught it.

The bold cells are what this fixes. "strips" means the key is gone; "blanks" means the runtypes `strip` decoder sets it to `undefined` and leaves it in place (its documented contract):

| shape | `huk` | `uke` | `ces` | `pjs` / `rjs` / `sj` | `cj` / `cjr` | `strip` decoder |
| --- | --- | --- | --- | --- | --- | --- |
| `{a: string}` | reports | reports | strips | strips | strips | blanks |
| `{a: string}[]` | reports | reports | strips | strips | strips | blanks |
| `[{a: string}, number]` | reports | reports | strips | strips | strips | **kept** |
| `{t: [{a: string}, number]}` | reports | reports | strips | strips | strips | **kept** |
| `{a: string}[] \| number` | **missed** | **missed** | strips | **leaked** | strips | **kept** |
| `[{a: string}, number] \| string` | **missed** | **missed** | strips | **leaked** | strips | **kept** |
| `{a: string} \| {b: number}` | reports | reports | n/a | strips | **decode kept** | blanks |
| `{a: string} \| Date` | reports | reports | n/a | strips | **decode kept** | blanks |

Plus: mion's `clone` strategy did not strip on ARRIVAL at all, for any shape.

## The fixes

1. **`clone` decodes with a stripping decoder.** A new compiled family, `rjs`, rebuilds each object from its declared shape on the way in. `mutate` and `direct` keep undeclared keys on arrival, by design.
2. **One conjunct narrows the shared union gate** (`atomicOnlyJsonIdentity`), so `pjs`, `rjs` and `sj` walk an array or tuple union member instead of handing it through.
3. **The strip pre-pass walks a tuple** instead of no-opping there.
4. **`hasUnknownKeys` / `unknownKeyErrors` / the strip pre-pass walk an atomic union member** (`unionAtomicMemberDescent`).
5. **The strip pre-pass's object arm gained the shape guard** the other three families already ran under; a JSON wire writes an absent optional as `null`.
6. **Compact decode uses the safe union restore**, matching its safe encode. A union's merged member stays keyed on the wire, so it has room for an undeclared key.
7. **Index signatures are open, for encoding and decoding.** A plain or template index signature admits every key; a key that matches no pattern is copied by the encoders and left alone by the decoders, and validation is the one place that refuses it. A union with an index-signature member keeps every key in every family, on both ends: no codec can tell a stray key on the object member from a key the record member declares. `cloneExactShape` keeps dropping a pattern miss, since a by-reference copy would break `clone(x) !== x`; the binary encoder still drops one, since it has no value type to write it with.

Each noop predicate moved with its emit. `rjs` delegates every non-rebuilding arm to `rj`, so the two decoders cannot drift on the arms they share. The direct encoder also swept once per index signature and wrote every key twice on a split-key record; it now sweeps once per object.

## Behaviour changes worth knowing

- A wrong-typed param on a `clone` route now answers a **validation error** instead of a serialization error. Same 422, more precise payload. Ten platform-adapter tests and two router tests re-pinned.
- `strictTypes` only checks params, and it can now fire only on a `mutate` or `direct` route: `clone` and `compact` drop the key while decoding, so there is nothing left to reject. The website says so.
- Key order inside a nested object in a union member is now declared order, not insertion order. Same JSON, different bytes.
- `pjs` no longer returns the caller's array by reference.
- A hand-written marker wrapper that injects `pjs` with `rj` now fails closed with a mismatched-families error; `clone` pairs `pjs` with `rjs`.
- `getValidationErrors` with `{checkUnknowns: true}` still answers `[{expected: 'union', path: []}]` for a union: it detects by branch failure. Pinned by a test.

## A pre-existing bug fixed along the way

`deserializeBodyParamsOrThrow` handed the FROZEN `EMPTY_PARAMS` sentinel to the decoder when no params arrived; a live decoder threw and reported a serialization error for what is really a missing body. It now returns the sentinel without decoding, pinned by a test that fails without the guard.

## The oracle that should have caught all of this

The value fuzz's walker refused to descend into a union, and no corpus type had an object inside an array or tuple member, so six unknown-key oracles ran green over these shapes while checking nothing.

- **O27** is a run-level coverage rule: a target whose TYPE carries a keyed shape anywhere must have been planted in at least once.
- **O26** asserts the stripping decoder DELETES a wire key rather than blanking it.
- The walker descends a union when exactly one member could have produced the value. The wire plant writes into every object of the encoded wire with no exceptions; the oracle then asks the type whether the container admits each surviving key (an index signature, `any`, `unknown`, a bare `object`, or a union arm that only runs on a value its member validates).
- Four new corpus targets, `AtomicOnlyUnion`, `ObjectInTupleSlot`, `RecordInArrayUnion` and `TupleWithRecordSlot`; every round-trip lane reports a wiring failure the same way.

## Tests

- Go emitter assertions per family and per shape, each A/B checked against its own reverted fix, plus the first table test `isNoopForRestoreJsonSafe` has ever had.
- `unknownKeyFamiliesAgree.test.ts` and `unionDecodeAgree.test.ts`: every family on the same value, with `toStrictEqual` for the deleting families and an explicit blank assertion for the strip decoder, so a blanked key can no longer pass as a deleted one.
- `stripRestoreShapes.test.ts`: `rjs` on a pattern-keyed record, a Map of objects and a registered class; every encoder and decoder keeps a pattern miss while `validate` refuses it.
- `ExtraParams.ts` gains a union case and a tuple case, so all ten strategy pairings see both.
- Deterministic unit rows for the walker's union descent, `containsKeyedShape`, `wireKeyAdmitted` and O27.

Green: `go test ./internal/... ./cmd/...`, `pnpm run test:ci`, `pnpm run test:bun`, `pnpm run lint`, `pnpm miondevx core fuzz value --quick`, `pnpm miondevx core fuzz roundtrip --quick`.

https://claude.ai/code/session_0116EDoeEy4UMWfPXgmbMtvE


container/website/content/01.rpc/02.server/01.routes.md
```diff
@@ -102,7 +102,7 @@ Declaring intent with `mion.query()` and `mion.mutation()` is recommended. It ma
 
 ## Strict Types
 
-By default, mion allows objects with extra properties to pass validation. You can enable `strictTypes` globally (in `createMionRouter`) or per route to reject objects with unknown properties. See the [Validation](/rpc/server/validation#strict-types) page for details.
+By default, mion allows objects with extra properties to pass validation. You can enable `strictTypes` globally (in `createMionRouter`) or per route to reject params with unknown properties. It applies to `mutate` and `direct` routes: `clone` and `compact` drop unknown keys while decoding, so there is nothing left to reject. See the [Validation](/rpc/server/validation#strict-types) page for details.
```
M-jerez: I think we need o be clear here we have to options, either hasuknow props for validation error, or the decoder strategies that removed unknow props

container/website/content/01.rpc/02.server/07.validation.md
```diff
@@ -48,7 +48,9 @@ mion provides two main validation functions with different use cases:
 
 ## Strict Types
 
-By default, mion allows objects with extra properties to pass validation. When `strictTypes` is enabled, objects with unknown or extra properties are rejected.
+By default, mion allows objects with extra properties to pass validation. When `strictTypes` is enabled, a param carrying an unknown property is rejected.
+
+In the router this applies to params on `mutate` and `direct` routes. `clone` and `compact` drop unknown keys while decoding, so there is nothing left to reject.
```
M-jerez: same here stric types and remove unknow props on decoding/encoding should be clearly separated

packages/router/src/security.spec.ts
```diff
@@ -85,11 +88,24 @@ describe('security: request body', () => {
   });
 
   it('a decode failure carries a fixed deserializeError, not the decoder text', async () => {
+    mion.initRoutes({echoBag});
+    // A prototype-named wire key is refused by the decoder itself, so this is the path
+    // where the compiled function's own message must not reach the client.
+    const response = await dispatch('/echoBag', '{"echoBag":[{"__proto__":"x"}]}');
```
M-jerez: I think , proto and constructor are allso disallowed by security, please pin all non supported props. 
also should/could we return are mode useful message other than 'Parameters might be of the wrong type' ?

we should propably also pin the behaviour, for clone or decode functions that are supposed to drop unknow as those should drop them even from records? or shoudl we still thorw for insecure props ? what woudl be the best behaviour in terms of security?

packages/run-types/test/features/indexSigNamedSibling.test.ts
M-jerez, line 1: please rename this to something inteligible

packages/run-types/test/features/prototypeKeys.test.ts
M-jerez, line 1: same here, this file is only cechking __proto__, while binary also prohivits, proto and constructor, 
please evaluate if `proto` adn `constructor` should actually be blocked. nad if so lets create a guidelines todo in this PR and will be solved separelly.
ofc this  shoudl also be applied to records etc.

packages/run-types/test/features/unknownKeyFamiliesAgree.test.ts
M-jerez, line 1: this file is missgin and union test , record test,  and obeyc | record, test to pin all different behaviours and make sure all functions works the same for those more complex scenarion

packages/run-types/test/fuzz/value/fuzz.integration.test.ts
```diff
@@ -215,6 +227,58 @@ const targets: FuzzTarget[] = [];
...
+  });
+}
+
+// --- target: an object hiding inside an ATOMIC member ---
+// An array and a tuple are atomic members of the flat union layout, so neither of these unions has
+// an object member at all, and every gate keyed on that handed the value through without compiling
+// the member. The object one level down is where the key hides. The tuple target covers the other
+// half of the same hole: the strip pre-pass used to no-op at a tuple node outright.
+{
+  const schema = RT.union([RT.array(RT.object({a: TF.string()})), TF.number()]);
```
M-jerez, line 244: any  reason why we using scehmas instead types here in this file, just curious if there is a reason?

packages/run-types/test/features/prototypeKeys.test.ts
M-jerez: All three names are already blocked, and this file already covers all three. Line 29:

```ts
const UNSAFE = ['__proto__', 'prototype', 'constructor'] as const;
```

It loops that list for the JSON decoders, for validation and for the binary road. The canonical list is shared, `reflection.UnsafePropertyNames` in `ts-go-runtypes/internal/reflection/unsafe_names.go:10`, and one generated guard serves every road, so JSON and binary cannot disagree.

On whether they *should* all be blocked, I measured the actual hazard against the exact shape our rebuild loops use (`r[k] = v[k]`):

```
r["__proto__"] = v     prototype CHANGED
r["constructor"] = v   plain own key, prototype untouched
r["prototype"] = v     plain own key, prototype untouched
```

So only `__proto__` is a real vector here. The other two are defensive, and they cost something: a `Record<string, string>` of form fields or tags cannot carry a key named `constructor` or `prototype`, the decoder throws, and those are ordinary words the declared type says are valid. It also contradicts what this PR settled elsewhere, that an index signature admits every key and validation is what refuses one.

Worth noting for `__proto__` too: `JSON.parse` already lands it as a safe own key, and `Object.defineProperty` writes it as a safe own key as well, so carrying it is an option rather than only throw or drop.

Filed as a guidelines todo in this PR, `refused-property-names-under-an-index-signature`, to be solved separately. It splits the question by position: a declared property keeps refusing all three, while under an index signature the key is data. Leaving this thread open for you.

---

packages/run-types/test/fuzz/value/fuzz.integration.test.ts
M-jerez: Because the oracles need the schema as a runtime value, not just a type.

A TypeScript type is gone at runtime. The fuzz walks the schema on every iteration: it generates values from it, decides where an undeclared key can be planted, asks whether the type carries a keyed shape anywhere, and now asks whether a container admits a given key. All of those take `target.schema` and traverse it, so it has to be a real object.

There is a second constraint that shows up if you try to tidy the file, and the header at the top records it: the plugin resolves each `createX` call statically from the type of its argument, so every factory has to be called against a concretely typed `const schema`. A generic helper taking a `RunType` parameter injects the `unknown` runtype instead. That is also why the targets are written out one by one rather than through a shared factory.

No code change here. Leaving the thread open for you to close.

---

container/website/content/01.rpc/02.server/01.routes.md
M-jerez: Rewritten. The section now names the two options up front and says what each one actually does: `strictTypes` rejects the request with a validation error, while the `clone` and `compact` params encoders rebuild the params from the declared type so the property is simply gone. A note states the gotcha plainly, that decoding runs before validation, so `strictTypes` only ever fires on a `mutate` or `direct` route. The detail stays on the Validation page, which the section still links.

Fixed in a5e31c1.

---

container/website/content/01.rpc/02.server/07.validation.md
M-jerez: Split into two bullets, reject against drop.

Reject is `strictTypes`: the param fails validation, the route never runs, the caller gets a `validation-error` (HTTP 422), and nothing is removed. Drop is the `clone` or `compact` params encoder: the params are rebuilt from the declared type while they are decoded, so the property is gone before validation and before the handler. A closing line says when you want each.

A note carries the ordering fact, decode before validate, so pair `strictTypes` with `mutate` or `direct`, plus the fact that `strictTypes` checks params only and never return values. Verified against `rejectUnknownKeysOrThrow` in the router rather than from the old docs.

While in the file I also fixed two pre-existing headings that carried a function name and a dash as punctuation, both of which the writing guidelines ban. Each section's first sentence already names its API, so nothing was lost, and no link pointed at those anchors.

Fixed in a5e31c1.

---

container/website/content/01.rpc/02.server/01.routes.md
```diff
@@ -102,7 +102,11 @@ Declaring intent with `mion.query()` and `mion.mutation()` is recommended. It ma
 
 ## Strict Types
 
-By default, mion allows objects with extra properties to pass validation. You can enable `strictTypes` globally (in `createMionRouter`) or per route to reject params with unknown properties. It applies to `mutate` and `direct` routes: `clone` and `compact` drop unknown keys while decoding, so there is nothing left to reject. See the [Validation](/rpc/server/validation#strict-types) page for details.
+There are two ways to deal with a param property your types do not declare, and they do different things. `strictTypes`, set in `createMionRouter` or on a single route, rejects the whole request with a validation error. The `clone` and `compact` params encoders instead rebuild the params from your declared type, so the property is gone before the route runs. See the [Validation](/rpc/server/validation#strict-types) page for details.
```
M-jerez: be more clear:
so unknown properties are droped at serialization instead validation.

please use similar text on all related updates if required

M-jerez: Done, and carried across the three related pages so they all use the same framing.

The opening line now names the stage outright:

> A param property your types do not declare is handled at serialization, not at validation. The `clone` and `compact` params encoders rebuild the params from your declared type while they decode the request, so the property is dropped before the route runs. That is the default.

Rejecting is now clearly the other thing, at the other stage, in its own sentence. The note says serialization runs before validation, which is why `strictTypes` only fires on a `mutate` or `direct` route.

Same wording on the other two:

- Validation page: two bullets, "Dropped at serialization" and "Rejected at validation", with a line on when you want each.
- Serialization page: the note now opens with "An unknown key is dropped at serialization, not at validation" and links to Strict Types for the reject side.

Fixed in fcc56d0.

---

packages/run-types/test/features/unknownKeyFamiliesAgree.test.ts
M-jerez: Added, three of them. What each one proves:

- **Union of two object shapes** (`{a: string} | {b: number}`): a key neither shape declares behaves exactly like the flat-object row. Reported, and removed by every function that removes.
- **Record** (`Record<string, number>`): the inverse answer. The index signature declares every key, so nothing is flagged and nothing is removed. A family that dropped a key here would be destroying the caller's data, which is why this one sits in its own test instead of the shared loop.
- **Record beside an object** (`Record<string, number> | {a: string}`): no codec can prove a key is stray, so every encoder and decoder keeps every key.

**One disagreement the third case exposed, and it is worth your call.** On `{"a":"x","evil":1}`:

```
hasUnknownKeys              false
unknownKeyErrors            []
validate                    true
validate {checkUnknowns}    FALSE
clone / direct / strip      keep every key
```

Everything agrees except the strict validator. It judges each member whole, and neither member fits cleanly: the record refuses the string `a`, and the object literal refuses the extra key. `hasUnknownKeys` instead reads the merged allowlist, where the record member declares every key, so nothing is undeclared. The test pins this with a comment naming the reason rather than bending an assertion, but if the rule is that a union with a record member has everything declared, then the strict validator is the family that did not get the memo. That is a codegen decision, not a test fix.

Also note `cloneExactShape` refuses an object-bearing union outright with CES001, thrown when the function is created rather than when it is called. Both union rows pin the refusal.

Fixed in d1c1be3.

---

packages/run-types/test/features/indexSigNamedSibling.test.ts
M-jerez: Renamed to `indexSigNamedSibling.test.ts`, matching the camelCase of `indexSigUnionDispatch.test.ts` next to it. Nothing in the repo referenced the old filename.

`G1` stays in the describe title, because it is a live fixture id used elsewhere: two repro rows in the type fuzz, and the Go codegen comments and tests. The title now reads `G1 index signature does not corrupt a named sibling property`, so the name says what it covers and the id is still greppable.

Fixed in d1c1be3.

---

packages/router/src/security.spec.ts
M-jerez: Three parts, one done and two answered.

**All three names are now pinned here.** The suite has a single `UNSAFE_KEYS` list, the canonical one, and both `it.each` sites read from it: each name planted as a wire key gets the same `serialization-error` and the same fixed `deserializeError`, with leak checks that neither the engine text nor the refused key itself reaches the client.

**A bug fell out while doing it.** The existing "prototype-named route id" test built its body with `JSON.stringify({__proto__: {x: 1}, ...})`. In an object literal `__proto__:` is the prototype setter, so that key never made it into the JSON:

```
JSON.stringify({__proto__: {x:1}, constructor: [1]})  ->  {"constructor":[1]}
```

The test had been proving nothing about `__proto__` for as long as it existed. It now builds raw JSON so all three names really reach the wire.

**On a better message than "Parameters might be of the wrong type".** Yes, and there is one clean candidate. The text is fixed because a compiled decoder's own message can quote internal detail, so the real one stays server-side on `originalError`. The unsafe-key refusal is the one case where that caution does not apply: the text is mion's own constant and it names a key the caller themselves sent, so it can be surfaced without leaking anything. It is also a bigger share of what reaches this path than it used to be, now that a wrong-typed param answers a validation error instead.

**On whether stripping functions should drop unsafe keys from records rather than throw.** I measured the actual hazard against the shape our rebuild loops use:

```
r["__proto__"] = v     prototype CHANGED
r["constructor"] = v   plain own key, prototype untouched
r["prototype"] = v     plain own key, prototype untouched
```

So only `__proto__` earns a refusal. For the other two, throwing rejects legitimate record data. And even `__proto__` has a third option: `Object.defineProperty` lands it as a safe own key, which is what `JSON.parse` already does.

Both of those are captured in the `refused-property-names-under-an-index-signature` todo filed in this PR, which splits the decision by position: a declared property keeps refusing all three, while under an index signature the key is data. Leaving this thread open for you.

Fixed in d1c1be3.

---

packages/run-types/test/features/unknownKeyFamiliesAgree.test.ts
M-jerez: Correcting myself on the previous reply: calling this a disagreement was wrong.

The two sides are asked different questions and both answer correctly. The unknown-key families ask whether any key is undeclared, and none is, because the record member declares every key. The strict validator asks whether the value matches a member strictly, and none does: the record refuses the string `a`, and the object literal refuses the extra key. There is no rule that the second is the first plus loose validation, and for a union it is not.

Nothing to change in the behaviour. I only reworded the note above the test so a future reader is not sent hunting for a bug that is not there.

Fixed in 207d57e.

---

PR #290, opened 2026-09-17
fix(runtypes): attribute findings per entry, and annotate the test suites instead of downgrading everything

M-jerez:
Three run-types vitest projects set `downgradeErrors: '*'`, which made a real finding indistinguishable from an expected one. That already cost us nine call sites that passed a non-literal encoder strategy and silently compiled the default instead, two of them security tests. `CTA001` named every one and nobody could see it.

Removing the wildcard turned up two resolver bugs, so this lands in four parts.

## 1. Findings were attributed to call sites that never asked for them

The spec behind this estimated "45 call sites across 15 suite files". A run with the wildcard off reported **7464 findings across 1517 lines in 85 files**. Almost all of it was mis-attribution, on two independent axes.

**Depth.** The 28 root-position codes are registered `ScopeRoot`, whose own comment says the same trigger inside a property is a different, child-position code. But provenance is inherited down the type graph (so a member-level warning reaches the site that pulled the member in), and the root codes rode that inheritance onto every site that merely *contained* the type:

```ts
// my code asks only to validate a string format...
createValidateFn<TF.String<{minLength: 2}>>()
// ...but reported this, because ANOTHER file had createJsonEncoderFn<never>():
//   error PJ001: Type `Never` can never be encoded to JSON: the generated
//   function will always fail.
```

One deliberate `createJsonEncoderFn<never>()` made 846 unrelated sites report that a function they never asked for always fails.

**Family.** Provenance was keyed by type id with no family, so a finding raised while rendering one family's entry reached every site that named the type. Only **19 of 333** `CES001` reports were actually clone sites; the rest were JSON, validate, binary and mock sites hearing a clone finding. A `createJsonEncoderFn` site reported `VL002`, and a validate site reported all six serializer root codes.

Same bug also meant the alwaysThrow runtime message, which takes `provenance[0]`, could name a random unrelated file.

Provenance is now keyed per rendered entry (type id + family tag) and kept in two maps: a `ScopeRoot` code reports at the sites that **named** the type, every other scope keeps the inherited sites so a child-position warning still reaches the site that pulled the member in. That took the tree from 7464 findings to **205 across 200 lines**.

## 2. A new `@mion-downgrade-error` directive

`@mion-expect-error` removes a finding, which is wrong when the finding is true and worth seeing and only the halt is unwanted. A suite pinning what a broken type does at runtime had no way to say that except build policy.

```ts
// @mion-downgrade-error VL002
export const alwaysThrows = createValidateFn<symbol>();
```

Same line-above placement and optional code list, but it keeps the finding and marks it downgraded on the wire. Level and severity stay whatever the catalog says, so the two consumers that decide whether to halt (the bundler plugin and `mion compile`) read that flag alongside their own `downgradeErrors` setting and print the same `(downgraded)` note.

Four self-checks, all warnings, mirroring the `EXP` family: `DWN001` lowered nothing, `DWN002` names a code the level table never lowers, `DWN003` a typo, `DWN004` a code that is already a warning and so was never halting anything. New lint rule `runtypes/invalid-downgrade-error`.

## 3. A forwarded build-time argument is not a value to read

Review caught that the `CTA` findings on our own source were not noise to annotate. They were not, and the cause was one thing: **`CompTimeArgs` means the build tracks that reference.** So there are only three honest answers, and a suppression comment is none of them.

**A runtime key has nothing to track**, and `rtUtils` already ships an untracked door for it (`getPureFnByKey` / `hasPureFnByKey`, which `entryTuple.ts:669` was already using). The registration path was going through the tracked call and then silencing the finding. `hasPureFn` becomes `hasPureFnByKey`, and `getCompiledPureFn` gains the untracked twin it was missing.

**A wrapper forwarding its own `CompTimeArgs` parameter** was a true false positive, and it leaked well past our source:

```ts
function myOptional<const F>(field: CompTimeArgs<F>) {
  return propMod({optional: true}, field);   // CTA001, LevelRuntimeError -> halts a strict build
}
```

The value cannot be at that site by construction. It arrives from the wrapper's own call sites, and each of those is walked and demanded at its outer `CompTimeArgs` position, so the literal is still required exactly once, where it is written. Judged on the parameter's written annotation through the same check the scan loop already uses to find these parameters, so no callee name is involved, matching the return-type rule beside it. The injection markers already bless this wrapper shape (`marker_wrapper_forward_test.go`).

Guards pinned in `comptimeargs_forward_test.go`: a plain `string` parameter, a `let` binding and a user's own type named `CompTimeArgs` all stay rejected. `CTA001` and `CTA003` now name the untracked twins in their detail text.

The library source ends with **zero** `CTA` findings and nothing suppressed, down from four suppressions.

## 4. The three configs

- `packages/run-types/vitest.config.ts` and `test/mock-format-isolation/vitest.config.ts` name **nothing at all** now. Every surviving finding says so in its own source: `@mion-downgrade-error` where it is true and worth reading (the alwaysThrow suites, the no-plugin pure-fn lanes, the shared-format-entry case), `@mion-expect-error` where it is noise.
- `vitest.converted.config.ts` keeps `['CES001']`, the one generated finding. `mion convert` copies the comments along with the source, so the directives carry over, but four `cloning/Unions.ts` cases write their union across several lines and convert collapses the whole type argument onto one, dropping the comment that was inside it.

Two annotations survive in the test tree, both genuinely deliberate: `RT.array(RT.boolean)` under a `@ts-expect-error`, and a registration with a `Math.random()` id testing the no-plugin throw. Registration has no untracked twin by design, since that is the reference the build exists to track.

## Verification

- `error CTA001` on a non-literal build-time argument **halts the run again**, probed with a `let` binding in a suite file. That is the spec's headline bar.
- No `EXP`/`DWN` self-check noise anywhere, so every directive is doing its job.
- `pnpm run test:ci`: all 23 vitest projects, 12,372 tests.
- `go -C ts-go-runtypes test ./internal/... ./cmd/...` green (42 packages), `pnpm run lint` and `pnpm miondevx core codegen all --check` green.
- The converted lane passes 7586 tests.

New Go coverage: `provenance_scope_test.go` (the "one level deeper" and "one family over" twins CLAUDE.md asks for, plus a paired `getRunTypeId` shape test), `downgradeerror_test.go` (the directive and all four self-checks), `comptimeargs_forward_test.go` (the forward plus its three guards). JS coverage extends `downgrade-errors.test.ts`.

Two existing expectations changed rather than regressed. `runtype-diagnostics.test.ts` wanted the encoder's `PJ011` reported at the decoder's line too; it is now on its own call only, which is the point of the family fix. And three pure-fn test lookups that pass a **literal** key inside a factory kept the tracked call, because the tracking is what the dependency-detection tests assert.

## Merge order

#289 has merged, so this no longer waits on anything.

## Labels

`website` (touches the docs content tree) and `pre-publish-e2e` (ships a new lint rule in `oxlint-recommended.json`, which consumers extend).

https://claude.ai/code/session_01Siyux8eeU24Ti9B9buv9rU


container/website/content/02.runtypes/08.diagnostics/01.error-levels.md
```diff
@@ -57,3 +57,26 @@ It works like TypeScript's `@ts-expect-error`. When the finding goes away, the u
 ::tip
 Always name the code. A bare comment covers anything reported on that line, and only your editor can then tell you it went stale, because the build does not run every check the editor does.
 ::
+
+## Keeping a Code but Not the Halt
```
M-jerez: not a good titel, please follow docs guidelines. alos this full section is to long, make it way soimpler adn shorther

docs/todos/registerformatpattern-value-raises-cta003.md
M-jerez, line 1: think this has been merged already, please cehck fi we need ot move to done or somethig

packages/run-types/src/runtypes/entryTuple.ts
```diff
@@ -789,6 +789,9 @@ function registerTypeFnTuple(utils: RTUtils, tuple: FnTypeTuple): boolean {
 // code is absent (the live createPureFn ships instead).
 function registerPureFnTuple(utils: RTUtils, tuple: PureFnTuple): boolean {
   const record = tupleToRecord<PureFnRecord>(PURE_FN_TUPLE_KEYS, tuple);
+  // The key comes off a tuple the emitter wrote, so it is read at runtime rather
+  // than at build time; nothing here is a consumer reference to track.
+  // @mion-expect-error CTA003
```
M-jerez: this seem like a valid usuage, maybe CTA003 is wong here, I think CTA003 is intended to be used only inside pure fucnions ? oplease double check as I might be worng , or maybe is intended to be used only for registrepurefnfactory or similar, but this seem like a legit case.

packages/run-types/src/runtypes/pureFn.ts
```diff
@@ -103,11 +103,16 @@ const HOLLOW_PLACEHOLDER: CompiledPureFunction = {
 function registerCore(caller: string, key: string, arg: unknown, wrap: boolean): CompiledPureFunction {
   if (isEntryTuple(arg)) {
     initFromTuple(arg as EntryTuple);
+    // Same runtime lookup as below: the key arrives as a parameter.
+    // @mion-expect-error CTA001
```
M-jerez: same here, is CTA001 intended to be used only inside pure fuctions? oplease double check as I might be worng

packages/run-types/src/runtypes/pureFn.ts
```diff
@@ -103,11 +103,16 @@ const HOLLOW_PLACEHOLDER: CompiledPureFunction = {
...
     const registered = getRTUtils().getCompiledPureFn(key);
     if (registered) return registered;
     // Fall through to the no-entry error below — a tuple that doesn't
     // register its own key is an emitter bug worth surfacing loudly.
   }
+  // `key` is this library's own runtime lookup, not a consumer reference the build
+  // could track, so there is no literal to read here.
+  // @mion-expect-error CTA001
```
M-jerez: same here

packages/run-types/src/builders/compose.ts
```diff
@@ -540,6 +540,9 @@ export function propMod<const M extends PropModifiers, const F>(
  *  (`key?:`) inside `object(...)`. The common modifier gets a terse spelling;
  *  reach for `propMod` for `readonly` or combinations. **/
 export function optional<const F>(field: CompTimeArgs<F>): PropModCarrier<{optional: true}, F> {
+  // A forward, not a call site: the scanner reflects the enclosing `object(...)`
+  // literal, so this argument is never read at build time.
+  // @mion-expect-error CTA001
```
M-jerez: same CTA001. issu as toerh places

packages/run-types/src/runtypes/entryTuple.ts
M-jerez: You were right, and so was the diagnostic. Fixed rather than annotated.

Not a scope rule though: the check runs on every call whose resolved signature has a `CompTimeArgs` parameter, with no pure-fn gate anywhere on the path. The actual rule is what the brand means, which is *the build tracks this reference*. `rtUtils` already ships an untracked door for a key that only exists at runtime, and `entryTuple.ts:669` in this very file was already using it:

```ts
// Runtime-key lookup — the UNTRACKED companion to usePureFn/getPureFn/hasPureFn.
// The `key` parameter is a plain `string` (NOT `CompTimeArgs<string>`) …
hasPureFnByKey(key: string): boolean
```

`record.key` comes off a tuple the emitter wrote, so there is nothing to track. Now `hasPureFnByKey`, suppression deleted.

---

packages/run-types/src/runtypes/pureFn.ts
M-jerez: Right again. Same cause as the `entryTuple` thread: `CompTimeArgs` means the build tracks that reference, and `key` here is whatever `registerCore` was handed, so there is nothing to track.

`getCompiledPureFn` had no untracked twin, so I added the one that was missing beside the two that already existed:

```ts
getCompiledPureFnByKey(key: string): CompiledPureFunction | undefined {
  return pureFnsCache[key];
}
```

Both sites now use it, suppressions deleted. Worth noting this is a *separate* answer from the `compose.ts` forward fix: these parameters are plain `string`, which the forward fix deliberately still rejects (pinned by `TestCompTimeArgsForward_PlainParamRejected`), so the narrowing there cannot quietly grow to cover these.

---

packages/run-types/src/runtypes/pureFn.ts
M-jerez: Same fix as the site above: `getCompiledPureFnByKey`, suppression deleted.

---

packages/run-types/src/builders/compose.ts
M-jerez: This one is a different cause from the pure-fn threads, and you were right that it is not a legitimate suppression: it is a true false positive, now **fixed in the resolver** rather than annotated.

`optional` forwards its own `CompTimeArgs` parameter into `propMod`'s. The const-trace reached a parameter and stopped, so it reported CTA001. But the value cannot be at that site by construction: it arrives from `optional`'s own call sites, and each of those is walked and demanded at its outer `CompTimeArgs` position. So the literal is still required exactly once, where it is actually written and where the author can fix it.

This also leaked past our source. Any consumer wrapping a builder hit it:

```ts
function myOptional<const F>(field: CompTimeArgs<F>) {
  return propMod({optional: true}, field);   // CTA001, LevelRuntimeError -> halts a strict build
}
```

Judged on the parameter's written annotation, through the same check the scan loop already uses to find these parameters, so no callee name is involved, matching the return-type rule beside it. The injection markers already bless this wrapper shape (`marker_wrapper_forward_test.go`), so this is symmetry rather than a new concession.

Guards in `comptimeargs_forward_test.go`: a plain `string` parameter, a `let` binding and a user's own type named `CompTimeArgs` all stay rejected. I also confirmed the accepting test fails without the fix, so it is a real regression test.

CTA001 and CTA003 now name the untracked twins in their detail text, so the next reader finds that door instead of a comment.

---

docs/todos/registerformatpattern-value-raises-cta003.md
M-jerez: Correct, #289 merged. Rebased onto current main and deleted this duplicate copy: the spec lives in `docs/done/` now, carried in by that PR.

The rebase also needed a real conflict resolved. `381351b` removed the `unsafeDeclaredMember` block my provenance commit had edited, so I took main's side; the function is gone upstream.

---

container/website/content/02.runtypes/08.diagnostics/01.error-levels.md
M-jerez: Rewritten. Retitled **"Downgrading a Code on One Line"**, which names the job plainly and reads as the sibling of "Expecting a Code on One Line" right above it, and cut from 22 lines to 14: one intro sentence, the snippet, one detail paragraph, a two-line tip. Dropped the sample output fence, since the section above it already shows the `(downgraded)` line format.

---

PR #293, opened 2026-09-18
feat(purefn)!: identify a pure function by its package and its body hash

M-jerez:
A pure function's id was `<package>/<path>#<name>`, and the name half came from one of two places depending on where the registrar call happened to sit. Half the ids were locations, half were content addresses:

```ts
export const slugify = registerPureFn(...);         // #slugify   a location
routes.orders.getById(inputFrom(o, (x) => x!.id));  // #Kq3f..    a content address
```

## The rule now

```
id = <package name> # CodeHash(the body as it ships)

@mionjs/run-types#Kq3f_xN9pQ2wLd
@acme/text#9Zt1bRm4cVaPqL
```

No path. Renaming the binding, renaming the export, or moving the file inside its package all leave the id alone. It changes when the function changes, so a stale reference is a clean miss and never a different body.

The package half stays because delivery reads ownership off it (`purefnids.IDPrefix`, `RUN_TYPES_ID_PREFIX`), both now `<package>#`.

### As it ships, not as written

The hashed body carries each dependency's id in place of the binding it was written with. Hashing the written form is cheaper, but two files in one package whose `helper` resolves to different imports have identical text and different behaviour, and the dedup would have silently shipped one body for both.

## What that cost

`id(A)` depends on `id(B)` for every dependency, so resolution became a memoised post-order walk. The governing rule: computing an id **is** producing the shipped body, one pass with one output.

- `resolve.go` holds `resolveCtx` and one memoised `entryFor(file, call)`. A helper referenced by ten dependents is extracted once per Program.
- The memo keys on the call node, not its position: `inputFrom(…).asArg()` and the inner `inputFrom(…)` start at the same offset.
- A dependency cycle has no answer, so it is now **PFE9015** (`LevelError`). Nothing rejected one before; an eager cycle recursed forever at materialisation, so this replaces a runtime hang with a build error.
- A `.d.ts` dependency's id is read off its branded literal type, which caps the recursion: a published dependency is a leaf.

## What this deletes

- **`BodyHash` everywhere.** It was `sha256(id + code)`, so once the id *is* the hash of the code it carries nothing. Gone from `Entry`, the emitted tuple (dropped in lock-step in `module.go` and `PURE_FN_REQUIRED_KEYS`), the 46 built-in rows, `types.ts`, the metadata wire and the client store.
- **PFE9004 entirely.** "Two pure functions share an id but have different bodies" is unreachable.

## Readable names still ride along

An id is a hash, which names nothing a reader can search for. `Entry.BindingName`, `protocol.PureFnSite.bindingName` and a generated `purefnids.NameOf(id)` carry the source name, so the built-in constants keep readable names (`IsDateStringYMD`, `isUUIDId`) with hash values.

## Three defects this surfaced

- **The disk cache had no drift check on its pure-fn edges.** A cached type entry stores the pure-fn ids its body reaches, and the field was documented as needing no check because the keys were stable strings. Content-addressed ids end that: edit a built-in body and its id moves while the type's structural id does not, so the entry still hit and handed back an `argsText` baking `utl.getPureFn('<retired id>')`. `AddMissingStubs` then degraded the validator to the family identity function, silently, on every build. Both ends now check against `purefnids.Has`.
- **An emitted local variable spelled a hash bare.** `const G5-z72czuSL8sC = …` is not valid JavaScript (base64url carries `-`). The alias table is keyed by the `purefnids` constants now, with an `identifierSafe` fallback.
- **`ModuleName` emitted a double slash** for an id with no package half, so a project without a package name wrote its mapper module to an unfindable path.

## Behaviour that changes for users

- Every pure-fn id in a build moves. Both ends regenerate together, so this is a rebuild, not a runtime break.
- The same helper written in two files of one package is now **one** entry, compiled and shipped once. Same for batches: two identical batches in two files share one batch id.
- Two identical bodies in **different** packages stay apart, because the package half differs.

## Verification

- `go -C ts-go-runtypes test ./internal/... ./cmd/...` green.
- `pnpm run test:ci`, all 7 batches green (488 test files).
- `pnpm run lint`, `pnpm run format`, `pnpm miondevx core codegen all --check` clean.
- Website docs updated (`02.runtypes/02.guide/09.pure-functions.md`), spec written up in `docs/done/`.

Labelled `pre-publish-e2e`: this changes public ids and package exports.

https://claude.ai/code/session_01N92Fyfr6nMvaX1WXNgQVMm

---
_Generated by [Claude Code](https://claude.ai/code/session_01N92Fyfr6nMvaX1WXNgQVMm)_


container/website/content/02.runtypes/02.guide/09.pure-functions.md
```diff
@@ -55,8 +55,21 @@ registerValidator(validate);
 registerValidator((v: unknown) => typeof v === 'string');
 ```
 
+## How a Helper Is Identified
```
M-jerez: remove this section, this are internal detail, and should be transparent to user.
as much add the commetns in one of the examples

packages/devtools/src/core/go-generated/diagnosticCatalog.generated.ts
```diff
@@ -962,6 +954,14 @@ export const DIAGNOSTIC_CATALOG: Record<string, DiagnosticEntry> = {
     detail:
       "A pure function's id is computed from where it lives: the package, the file\nand the name it is bound to. The build injects it, so source normally passes\nnone. An id written by hand, or left behind by a move or a rename, would\nregister the body under one id while every reference to it uses the other.\n\nFix: delete the argument and let the build inject it, or regenerate the file\nthe id is imported from.",
   },
+  PFE9015: {
+    headline: 'Pure function `{0}` and the one reaching it here depend on each other; neither can be given an id.',
```
M-jerez: lest just log something like pure functions circular dependency, and lets list the two files that generates the circular dep

container/website/content/02.runtypes/02.guide/09.pure-functions.md
M-jerez: Section removed. The three points moved into the `custom-pure-fn.ts` example as a comment, which also fixes a real bug there: that comment still spelled the retired id shape (`'@acme/text/src/guide/custom-pure-fn#slugify'`).

```ts
// The registrar gives it back its id: your package name plus a short hash of
// the helper itself, like '@acme/text#9Zt1bRm4cVaPqL'. Rename it, rename what
// you export it as, or move it to another file in the same package and the id
// stays put. Edit the helper and you get a new id, because it is a new helper.
```

---

packages/devtools/src/core/go-generated/diagnosticCatalog.generated.ts
M-jerez: Done. Both files now ride in the message:

```
Pure functions circular dependency: `{0}` (`{1}`) reaches back into `{2}`.
```

`{0}` is the helper reached, `{1}` the file it is declared in, `{2}` the file holding the reference back into it. Same file at both ends when the cycle is written inside one module, which still reads correctly.

Threading it out meant `registrationIDOfBinding` returning the declaring file alongside the cycle flag.

While adding it I found PFE9015 had **no test at all**, so there are two now, across files and within one file. Each also stands in for "the extraction terminates": if the in-progress guard regresses they hang instead of failing.

---

PR #335, opened 2026-09-22
fix: host the sechttp fuzz suite in its own package and break the tsconfig reference cycle

M-jerez:
## What

`tsc --build` could not run over this workspace at all. Eight cycles in the project-reference graph, all through one edge: `packages/router/tsconfig.json` referenced `../test-server`.

```
$ pnpm exec tsc --build tsconfig.json --dry
error TS6202: Project references may not form a circular graph. Cycle detected:
  tsconfig.json
  packages/router/tsconfig.json
  packages/test-server/tsconfig.json
```

pnpm warned about it on every install too.

## Why the edge existed

One file: `packages/router/test/fuzz/security/httpFuzzRunner.ts`, the `sechttp` fuzz lane. It imported `@mionjs/test-server` and reached into `platform-node`'s source by relative path. The second one was only legal because the test-server reference gave tsc a transitive path to platform-node, so deleting the reference alone produced TS6059 and TS6307.

## What changed

**The suite moved into a new private package, `@mionjs/test-router-fuzz`.** It needs router, the node adapter and the test-server fixture at once. Any package holding all three ends up referenced by one of them, which closes the cycle again; a private package that nothing references cannot. `packages/type-budget` is already this exact shape, so this copies a convention rather than inventing one.

```
packages/test-router-fuzz/package.json
packages/test-router-fuzz/tsconfig.json
packages/test-router-fuzz/vitest.config.ts
packages/test-router-fuzz/CLAUDE.md
packages/test-router-fuzz/test/fuzz/security/{httpFuzz.integration.test.ts,httpFuzzRunner.ts}
```

Router's `test/` held nothing else, so the directory, the tsconfig reference and the `@mionjs/test-server` devDependency are all gone. Every cross-package import is a plain package name now:

```ts
import {createMionRouter, ..., getRouterFatalErrorResponse} from '@mionjs/router';
import {setNodeHttpOpts, startNodeServer, resetNodeHttpOpts} from '@mionjs/platform-node';
import {compactTestRoutes} from '@mionjs/test-server';
```

`getRouterFatalErrorResponse` was a dynamic `import()` inside a catch block, so it failed at run time rather than type-check time until it was made static.

**The package has one tsconfig and no `references`,** so it cannot close a cycle: nothing points at it and it points nowhere. `composite: false` + `rootDir: "../.."` is what keeps the relative import of the run-types fuzz core legal (a composite project with `rootDir: "."` refuses a source file above its root, TS6059). Its vitest config does install `mionVitePlugin`, unlike the other test-only projects: the suite declares its own fixture routes, and without build-time type information every one fails with `MissingRtFnsError`.

**`packages/test-server` is untouched.** An earlier revision of this PR hosted the suite there, which forced it into a typecheck-config plus build-config split. That shape only `drizzle-orm-pg-core` needs, and test-server ships, so it was reverted. `git diff origin/main -- packages/test-server` is empty.

**Registration:** the vitest project list, the `mion-core` batch (where the suite already ran as part of `router`), the heavy/light contract, and the two eslint brace globs in the root `package.json`. Private packages are filtered out of every release script and of the thin-README and publishable-manifest contracts, and `scripts/ci/lanes.mjs` classifies by the `packages/` prefix, so nothing else needed an entry.

**A gate stops the cycle coming back.** `scripts/ci/check-tree.mjs` gained a sweep over the project graph, unit-tested in `repo-contracts.test.ts` with fixtures plus one case over the real tree. Putting the old reference back fails it:

```
==> no tsconfig project reference cycle — 4 offender(s):
  packages/router/tsconfig.json -> packages/test-server/tsconfig.json -> packages/router/tsconfig.json
  packages/platform-node/tsconfig.json -> packages/router/tsconfig.json -> ... -> packages/platform-node/tsconfig.json
  packages/platform-cloudflare/tsconfig.json -> ...
  packages/platform-vercel/tsconfig.json -> ...
```

The reasoning lives in the repo, not only in the spec: `packages/test-router-fuzz/CLAUDE.md` says why the package exists and why its tsconfig carries no references, and `packages/router/CLAUDE.md` carries the rule that a package's test tree never imports a downstream consumer.

## Not in scope

This does **not** let the per-package `tsconfig.test.json` files go away. `tsc --build --noEmit` is refused outright with 45 instances of `error TS6310: Referenced project '...' may not disable emit`, cycle or no cycle. The reason to do this work is a correct project graph and a test file that stops reaching into two other packages illegally.

The `devtools` / `run-types` package.json cycle is untouched: different mechanism, no tsconfig references, and it does not block build mode.

## Verified

- `pnpm exec tsc --build tsconfig.json --dry` runs with zero TS6202.
- `pnpm miondevx core fuzz sechttp --quick` passes, all six oracles intact (SH-ALIVE, SH-ENVELOPE, SH-NO5XX, SH-NOLEAK, SH-TIME, SH-PROTO).
- `pnpm run test:ci` green across all 24 projects.
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run check:test-batches`, `pnpm run check-format` all green.
- Comment simplification pass ran (`chore(comments)`). No website page or example changed, so no docs pass was needed.

https://claude.ai/code/session_01N5Z2A4LhjmGu8YCsTpsGRQ


M-jerez:
### `js tests + lint` went red on an earlier head, and it is not this PR's

Job [106952264884](https://github.com/MionKit/mion/actions/runs/35788800071/job/106952264884) failed on `6a88886`, in the `test:bun` step:

```
##[group]src/bunHttpMiddleware.test.ts:

# Unhandled error between tests
-------------------------------
error: createMionRouter has already been called: create the router once per app (resetRouter() clears it in tests)
      at createMionRouter (packages/router/src/router.ts:192:5)
      at <anonymous> (packages/platform-bun/src/bunHttpMiddleware.test.ts:24:16)

 12 pass
 0 fail
 1 error
```

**Why it is not this PR's.** `git diff origin/main -- packages/platform-bun` is empty, and the two files involved are untouched here. `pnpm run test:bun` passes on this branch locally (`15 pass, 0 fail`), which is the tell: the failure is order-dependent.

**The mechanism.** Both bun test files call `createMionRouter` in their `describe` body and only call `resetRouter()` in `beforeAll`:

```
packages/platform-bun/src/bunHttp.test.ts:34            const mion = createMionRouter({...});  // describe body
packages/platform-bun/src/bunHttpMiddleware.test.ts:23  const mion = createMionRouter({...});  // describe body
packages/platform-bun/src/bunHttpMiddleware.test.ts:32    resetRouter();                       // only in beforeAll
```

Bun runs both files in one process and evaluates every describe body before any hook, so whichever body is evaluated second hits the module-level once-guard and throws. That file's tests then never run, and the summary reads `0 fail, 1 error`, which looks healthy at a glance.

**What is happening about it.** Delegated to a parallel session with its own spec and its own PR, per the repo's findings rule. No fix exists yet to port into this branch, so nothing was changed here. The once-guard itself is correct and stays.

**Merge order:** that fix merges before this PR.

---

